PR #27, opened 2022-03-09
Create ReadMe_CN中文.md

hehanbo1990:
js-tableland CHN translation....


hehanbo1990:
> @hehanbo1990 Thanks for this. I think we can merge, but @carsonfarmer and I were discussing making [some changes](https://github.com/tablelandnetwork/js-tableland/issues/29) to the EN readme today. Having something in CHN is better than nothing, but would you be willing to revisit after those changes? Thanks again, this is much appreciated!

surely!my pleasure!

PR #31, opened 2022-03-10
Joe/structure

### LINKED ISSUE #30: add structure to response from `create`
expose changes from https://github.com/textileio/go-tableland/issues/66 to consumer

joewagner:
Fixes #30

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -72,6 +72,10 @@ export interface CreateTableReceipt {
   description?: string;
 }
 
+export interface StructureHashReceipt {
+  structureHash: string;
+}
+
 export interface TableRegistrationReceipt {
   receipt: ContractReceipt;
   tableId: BigNumber;
@@ -94,4 +98,5 @@ export interface Connection {
     options: CreateTableOptions
   ) => Promise<CreateTableReceipt>;
   query: (query: string) => Promise<null | ReadQueryResult>;
+  hash: (query: string) => Promise<StructureHashReceipt>;
 }
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -47,6 +47,11 @@ export interface Row extends Array<any> {
   [index: number]: string | number;
 }
 
+export interface KeyVal {
+  0: string;
+  1: any;
+}
+
 export interface ReadQueryResult {
   columns: Array<Column>;
   rows: Array<Row>;
@@ -63,7 +68,7 @@ export interface CreateTableOptions {
 
 export interface CreateTableReceipt {
   name: string;
-  id: string;
+  structureHash: string;
   description?: string;
 }
 
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -66,10 +66,7 @@ export interface Row extends Array<any> {
   [index: number]: string | number;
 }
 
-export interface KeyVal {
-  0: string;
-  1: any;
-}
+export type KeyVal<T = any> = [string, T];
 
 export interface ReadQueryResult {
   columns: Array<Column>;
@@ -100,10 +97,10 @@ export interface TableRegistrationReceipt {
   tableId: BigNumber;
 }
 
-export interface RpcReceipt {
+export interface RpcReceipt<T = any> {
   jsonrpc: string;
   id: number;
-  result: any;
+  result: T;
 }
 
 export interface Connection {
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -22,6 +22,25 @@ export interface ConnectionOptions {
   network?: string;
 }
 
+export interface RpcParams {
+  controller?: boolean;
+  createStatement?: string;
+  description?: string;
+  dryrun?: boolean;
+  statement?: string;
+  tableId?: string;
+}
+
+export interface RpcRequestParam {
+  controller?: string;
+  /* eslint-disable-next-line camelcase */
+  create_statement?: string;
+  description?: string;
+  dryrun?: boolean;
+  id?: string;
+  statement?: string;
+}
+
 export interface ConnectionReceipt {
   jwsToken: Token;
   ethAccounts: Array<string>;
--- a/src/lib/tableland-calls.ts
+++ b/src/lib/tableland-calls.ts
@@ -1,8 +1,10 @@
 /* eslint-disable node/no-missing-import */
 
+import camelCase from "camelcase";
 import {
   TableMetadata,
   ReadQueryResult,
+  KeyVal,
   CreateTableOptions,
   CreateTableReceipt,
   Connection,
@@ -29,6 +31,18 @@ async function sendResponse(res: any) {
   if (json.error) throw new Error(json.error.message);
 
   return json.result;
+}
+
+// Take an Object with any symantic for key naming and return a new Object with keys that are lowerCamelCase
+// Example: `camelCaseKeys({structure_hash: "123"})` returns `{structureHash: "123"}`
+function camelCaseKeys(obj: object) {
+  return Object.fromEntries(
+    Object.entries(obj).map((entry: KeyVal) => {
+      const key = entry[0];
+      const val = entry[1];
+      return [camelCase(key), val];
+    })
+  );
 }
 
 async function GeneralizedRPC(
@@ -85,7 +99,7 @@ export async function create(
   const response = await SendCall.call(this, message);
   const json = await sendResponse(response);
 
-  return json;
+  return camelCaseKeys(json) as CreateTableReceipt;
 }
 
 async function query(
@@ -96,7 +110,7 @@ async function query(
   const response = await SendCall.call(this, message);
   const json = await sendResponse(response);
 
-  return json;
+  return camelCaseKeys(json) as ReadQueryResult;
 }
 
 export { query, list, TableMetadata };
--- a/src/lib/tableland-calls.ts
+++ b/src/lib/tableland-calls.ts
@@ -53,19 +53,17 @@ async function GeneralizedRPC(
   method: string,
   params: RpcParams = {}
 ) {
+  const signer = this.signer;
+  const address = await signer.getAddress();
+
   const param: RpcRequestParam = {
+    controller: address,
     create_statement: params.createStatement,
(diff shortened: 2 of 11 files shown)


src/lib/tableland-calls.ts
```diff
@@ -31,38 +36,49 @@ async function sendResponse(res: any) {
...
 ) {
-  const signer = this.signer;
-  const address = await signer.getAddress();
-
-  const params = [
-    {
-      statement: statement,
-      id: tableId,
-      controller: address,
-      type: options?.type,
```
joewagner, line 49: As far as I can tell`type` is not being used.  Maybe an artifact from an older RPC spec?

src/lib/tableland-calls.ts
```diff
@@ -31,38 +36,49 @@ async function sendResponse(res: any) {
...
-  ];
+  const param: RpcRequestParam = {
+    create_statement: params.createStatement,
+    description: params.description,
+    dryrun: params.dryrun,
+    id: params.tableId,
+    statement: params.statement,
+  };
+
+  if (params.controller) {
```
joewagner, line 64: Only some of the Validator RPC endpoints are expecting `controller`.  I was thinking that instead of always sending it, this function would be called with a boolean that indicated if it should be included.  Open to other ideas though

src/lib/tableland-calls.ts
```diff
@@ -31,38 +36,49 @@ async function sendResponse(res: any) {
   return json.result;
 }
 
+// Take an Object with any symantic for key naming and return a new Object with keys that are lowerCamelCase
+// Example: `camelCaseKeys({structure_hash: "123"})` returns `{structureHash: "123"}`
+function camelCaseKeys(obj: object) {
```
joewagner, line 41: The RPC results body has the convention of snake_case, and we want the convention of camelCase.

test/structureHash.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    const createStatement = "CREATE TABLE hello (id int primary key, val text);";
+    const hashResponse = await connection.hash(createStatement);
+
+    const payload = JSON.parse(fetch.mock.calls[0][1]?.body as string);
+
+    // test that faux response makes it through
+    await expect(hashResponse.structureHash).toEqual("ef7be01282ea97380e4d3bbcba6774cbc7242c46ee51b7e611f1efdfa3623e53");
+
+    // test that fetch is called how validator expects
+    await expect(payload.params[0]?.create_statement).toEqual(createStatement);
```
joewagner, line 36: Our tests aren't really testing what the Validator responds with, we would need end-to-end test for that.  My thought here is that this spy will at least ensure that mock response is mapped to the expected SDK's return values.  i.e. snake_case => camelCase.

joewagner (commented):
I did a little bit of refactoring on how the RPC requests are built and how the responses are parsed.  A detailed review might be in order.

src/lib/tableland-calls.ts
carsonfarmer: Do any of them actually take it anymore?

src/lib/tableland-calls.ts
```diff
@@ -75,28 +91,45 @@ export async function create(
...
+    statement: query,
+    dryrun: options.dryrun,
+    controller: true,
+  });
+
   const response = await SendCall.call(this, message);
   const json = await sendResponse(response);
 
-  return json;
+  return camelCaseKeys(json) as CreateTableReceipt;
```
carsonfarmer, line 104: Is it possible to specify CreateTableReceipt from the result of sendResponse? That is a "safer" way if sendResponse is generic.

src/lib/tableland-calls.ts
awmuncy: This is correct.

test/structureHash.test.ts
carsonfarmer: Yeh this is a good addition. I'd really like to have an easy way to actually spin up a server instance and test against "live" response types... that might be something we can get to by end of the next season.

src/lib/tableland-calls.ts
carsonfarmer: Is this just because JS likes camelCase?

test/structureHash.test.ts
awmuncy: That makes sense. I would like to figure out end-to-end tests, but for now I think this works.

src/interfaces.ts
```diff
@@ -47,6 +66,11 @@ export interface Row extends Array<any> {
   [index: number]: string | number;
 }
 
+export interface KeyVal {
```
carsonfarmer, line 69: I think maybe a tuple type here might be nicer? And while we're at it, we might as well make it generic rather than using any?

```typescript
```

This allows us to do things like the following, which will also catch things like trying to index out of range etc.

```typescript
const one: KeyVal<string> = ["a", "b"];
console.log(one[0]);
```

src/lib/tableland-calls.ts
awmuncy: The way you've done this is quite clean, so I like it. I suspect this will make working on the SDK easier, but it's also possible it will trip folks up; perhaps one might read the RPC response structure, and unless one happens to know the  case is transformed, find it more difficult to map those things in when reading the SDK. Occasionally when debugging, I find a simple "CTRL + F" in the relevant codebase helps, and this means that wouldn't yield results in some cases.

src/interfaces.ts
carsonfarmer: We _should_ also make `RpcReceipt` generic, because we'll know exactly the response type we're expecting given the method we're calling:

```typescript
export interface RpcReceipt<T = any> {
  jsonrpc: string;
  id: number;
  result: T;
}
```

Note also that this simplifies some of the types:

```typescript
export type Column = Array<ColumnDescriptor>;

export type Row = Array<string | number>;
```

src/lib/tableland-calls.ts
```diff
@@ -31,38 +36,49 @@ async function sendResponse(res: any) {
   return json.result;
```
awmuncy, line 36: Could we put the 'camelCaseKeys' function here. and remove it from the other functions? It looks like anywhere "SendCall" is used, the response is wrapped in `camelCaseKeys` anyway.

src/lib/tableland-calls.ts
carsonfarmer: Yeh nice catch, I think you're right.

awmuncy (changes requested):

carsonfarmer (changes requested):
This looks great @joewagner. I've made a few comments, and also added some thoughts about types. To date, we've made limited use of proper types here, and so I have kinda hijacked your PR here to suggest we make better use of type safety. What do you think, get those updates in here, or do as a separate PR?

src/lib/tableland-calls.ts
joewagner: I think so, although I didn't spend any time double checking.  [tableland-openapi-spec.yaml](https://github.com/textileio/go-tableland/blob/main/tableland-openapi-spec.yaml) should have the params that are currently expected.  tableland_createTable, and tableland_runSql both have it in their spec

src/lib/tableland-calls.ts
joewagner: > Is this just because JS likes camelCase?

I didn't think to much on this because our current linting wants everything lowerCamel.  In general I think the Javascript community prefers object properties to be lowerCamel, myself included, but I know a lot of libraries don't do this.  I don't have a strong opinion, as long as we stay consistent.

> The way you've done this is quite clean, so I like it. I suspect this will make working on the SDK easier, but it's also possible it will trip folks up; perhaps one might read the RPC response structure, and unless one happens to know the case is transformed, find it more difficult to map those things in when reading the SDK. Occasionally when debugging, I find a simple "CTRL + F" in the relevant codebase helps, and this means that wouldn't yield results in some cases.

This is a good point.  As mentioned above, I have a slight preference for lowerCamel, but it's definitely just a preference.

test/structureHash.test.ts
joewagner: Excellent!  I've been thinking about end to end, and I don't think it would be that complicated.  Happy to tackle that task, or help out.  I'll admit I'm not exactly sure how that type of task will map to a KR, but I'm sure it will.

src/lib/tableland-calls.ts
joewagner: Good call!

joewagner:
> To date, we've made limited use of proper types here, and so I have kinda hijacked your PR here to suggest we make better use of type safety. What do you think, get those updates in here, or do as a separate PR?

Really good points!  I thought about using the tuple, but wasn't sure.  I'm admittedly not the most up to date on typescript best practice.  I like the idea of catching attempts to index out of range.  Maybe a separate PR does make sense?

joewagner:
Thanks for all the feedback, much appreciated!  Just realized I left the PR as a draft, updated now.  I'll make the requested changes and push again shortly.

src/lib/tableland-calls.ts
joewagner: I may have made a bad assumption here... Looks like [tableland-openapi-spec.yaml](https://github.com/textileio/go-tableland/blob/main/tableland-openapi-spec.yaml) is not accurate.  Gonna put the controller back on to all requests.

awmuncy (approved):

src/lib/tableland-calls.ts
carsonfarmer: But I'm not sure we need controller at all anymore. Can you confirm if this is needed at all? Additionally, we should really make sure we do update that spec, it looks like @jsign has recently updated it, so maybe it is the right source of truth?

carsonfarmer (approved):

src/lib/tableland-calls.ts
joewagner: I can confirm we **do** need the controller.  If I remove it, the Validator returns an RPC error.  Maybe a different issue could be opened in `go-tableland` to update the spec to include it where it's needed, and remove it elsewhere?

PR #35, opened 2022-04-18
Fix #34 by passing all options

### LINKED ISSUE #34: The description option no longer works
## Expected Behavior
When using the SDK to create a table, the caller used to be able to add a description to the table via an options Object

## Actual Behavior
The description is ignored

## Steps to Reproduce the Problem

1. Call the `create` function with an options object containing a description property
2. Call the `list` function with the same wallet as step one and notice that no description exists

joewagner:
Fixes #34


carsonfarmer (approved):
👍

PR #37, opened 2022-04-20
Change Bearer token to encoded SIWE message

joewagner:
This update replaces the JWT Bearer token with a JSON object containing a SIWE message and a signature that is then base64 encoded.

This update must not be published until https://github.com/textileio/go-tableland/pull/86 has been deployed.

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -1,0 +1,18 @@
+declare let globalThis: any;
+
+// From https://www.npmjs.com/package/btoa
+const ployfills = {
+  btoa: function (str: string | Buffer) {
+    let buffer;
+
+    if (str instanceof Buffer) {
+      buffer = str;
+    } else {
+      buffer = Buffer.from(str.toString(), "binary");
+    }
+
+    return buffer.toString("base64");
+  },
+};
+
+export const btoa = globalThis.btoa ?? ployfills.btoa;
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -1,7 +1,7 @@
 declare let globalThis: any;
 
 // From https://www.npmjs.com/package/btoa
-const ployfills = {
+const polyfills = {
   btoa: function (str: string | Buffer) {
     let buffer;
 
@@ -15,4 +15,4 @@ const ployfills = {
   },
 };
 
-export const btoa = globalThis.btoa ?? ployfills.btoa;
+export const btoa = globalThis.btoa ?? polyfills.btoa;
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,6 +40,7 @@
         "jest-fetch-mock": "^3.0.3",
         "jest-ts-webcompat-resolver": "^1.0.0",
         "prettier": "^2.5.0",
+        "siwe": "^1.1.6",
         "ts-jest": "^27.1.3",
         "ts-node": "^9.1.1",
         "typedoc": "^0.22.0",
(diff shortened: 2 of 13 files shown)


joewagner:
Looks like the Github Action is using an old version of Node, in which the `btoa` global isn't supported.
@carsonfarmer @awmuncy Do we want to use a polyfill?

joewagner:
> Do we want to use a polyfill?

A few alternatives:
 - we could require that anyone using this lib has `btoa` available globally
 - we could limit Node.js support to versions >= 16

carsonfarmer:
No those requirements are too restrictive. I already don't like requiring fetch as a global, its a rough DX. Use this https://www.npmjs.com/package/@stablelib/base64. Its a super lightweight dep.

carsonfarmer:
Requiring new node versions is rough for folks running in restricted envs (like GitHub Actions, or edge workers etc). I'm even re-evaluating if we should include a fetch ponyfill...

joewagner:
> Requiring new node versions is rough for folks running in restricted envs (like GitHub Actions, or edge workers etc). I'm even re-evaluating if we should include a fetch ponyfill...

If we package node-fetch into the SDK it would mean adding 106 kB. I'm ok with that.
@carsonfarmer you ok if I add the btoa polyfill to this PR and we do the fetch polyfill in a separate one?  
I'm thinking splitting them would help move this along and keep the PR from getting too big

joewagner:
Just pushed with  a polyfill for `btoa`, but having rebased onto main I found an issue with the way we are publishing the source typescript files. I'll add inline comments describing.

Related: https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-import.md

src/lib/util.ts
```diff
@@ -0,0 +1,18 @@
...
+
+// From https://www.npmjs.com/package/btoa
+const ployfills = {
+  btoa: function (str: string | Buffer) {
+    let buffer;
+
+    if (str instanceof Buffer) {
+      buffer = str;
+    } else {
+      buffer = Buffer.from(str.toString(), "binary");
```
carsonfarmer, line 11: Is it safe to use this? Is this going to require Buffer in the browser now? Or will this be tree-shaken out?

src/lib/util.ts
```diff
@@ -0,0 +1,18 @@
+declare let globalThis: any;
+
+// From https://www.npmjs.com/package/btoa
+const ployfills = {
```
awmuncy, line 4: It's a ploy! Should be `polyfills` not `ployfills`.

joewagner:
After chatting with @jsign I am changing the base for this PR to a v2 branch in the hopes that we can keep that branch working with the changes happening in https://github.com/textileio/go-tableland/tree/v2

src/lib/util.ts
carsonfarmer: Just making sure we understand the implications here? Is this going to be an issue with our browser builds? I don't want to bundle Buffer if we don't have to (or aren't already).

carsonfarmer (approved):
One q otherwise looks good.

awmuncy (approved):

src/lib/util.ts
awmuncy: My understanding: We won't be bundling Buffer, but expecting it to already be in the global environment in which this code is called. In modern contexts in which a btoa polyfil is needed (node environments), Buffer will already be built in. In environments where Buffer isn't built in (browsers), this line of code will never be hit because btoa is built in.

carsonfarmer: Well the `if (str instance Buffer)` line will be hit every time this runs, and in my browser if I run that line, I see:
```
> "" instanceof Buffer
< ReferenceError: Can't find variable: Buffer
```

So this seems unsafe to me.

carsonfarmer: Unless `ployfills` is only _included_ when we need it? In which case, that seems fine!

PR #40, opened 2022-04-20
Joe/response types

### LINKED ISSUE #36: Incorrect types for query responses
It looks like the Typescript types for query responses are currently incorrect. Whereas I would expect, based on the types that the following returns an object with `columns` and `rows`, I actually get an object with `{ data: { columns, rows } }`. However, the Typescript compiler complains with `Property 'data' does not exist on type 'ReadQueryResult'.`. The fix is to either automatically destructure the returned object (internally) and return rows and columns, or to update the types to reflect the fact that the data is structured under the data key. The former seems like the nicer option.

```
const response = await connection.query(`select * from some_table;`);
```

### LINKED ISSUE #38: Linting fails after publishing src/
As a fix for https://github.com/tablelandnetwork/js-tableland/issues/33 we published the typescript source. This introduced an new issue because the source in now in violation of the eslint rule `node/no-unpublished-import` ref: https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-import.md

A solution that should meet all the requirements is to add eslint disables for that rule.

joewagner:
Fixes #36

The successful RPC response data has a few different schema depending on the endpoint, and on how the endpoint is called.  Keeping in mind that some of this will change as this season's changes to how we interact with Smart Contracts start to be implemented.

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -69,8 +69,10 @@ export interface Row extends Array<any> {
 export type KeyVal<T = any> = [string, T];
 
 export interface ReadQueryResult {
-  columns: Array<Column>;
-  rows: Array<Row>;
+  data: {
+    columns: Array<Column>;
+    rows: Array<Row>;
+  };
 }
 
 export interface CreateTableOptions {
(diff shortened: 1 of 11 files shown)


carsonfarmer (approved):
Nice! Our Typescript friends will be pleased.

PR #43, opened 2022-05-05
Add Bearer token to requests to list an addresses tables

joewagner:
This adds an Authorization header to http requests to `/tables/controller/${address}`.  This matches what is expected by the v2 branch of the Validator.
We are sending the SIWE token to this route now because the Validator has become multi-chain and it will use the decoded SIWE message to determine the chainId.

### DIFF
--- a/src/lib/list.ts
+++ b/src/lib/list.ts
@@ -4,7 +4,12 @@ export async function list(this: Connection): Promise<TableMetadata[]> {
   const address = await this.signer.getAddress();
 
   const resp: TableMetadata[] = await fetch(
-    `${this.host}/tables/controller/${address}`
+    `${this.host}/tables/controller/${address}`,
+    {
+      headers: {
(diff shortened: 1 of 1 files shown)


carsonfarmer (approved):
I'm guessing those errors are due to un-released contract client?

joewagner:
> I'm guessing those errors are due to un-released contract client?

@carson I think thats correct.  Once https://github.com/tablelandnetwork/eth-tableland/pull/11 is merged I believe we could publish a v2 tag of eth-tableland to npm and then update this repo's v2 branch dependancy to "@tableland/sdk": "v2"
Maybe call the tag `latest` instead of `v2`?  I don't have strong preference

joewagner:
After some discussion with Bruno and Ignacio, the validator is no longer going to require the Authorization header for this route.  Which means no need to merge this, closing.

PR #44, opened 2022-05-06
Export a function called `receipt` that calls the Validator to get transaction receipt

joewagner:
The Validator has a new rpc endpoint `tableland_getReceipt` which returns information about the specified transaction hash.  
This will allow verifying if a Smart Contract transaction was processed by the Validator.
Although this PR does not implement the functionality, we could now ensure that the Validator the SDK has connected to has finished processing a create table transaction before the SDK returns from a call to the `create` function


src/interfaces.ts
```diff
@@ -70,6 +73,16 @@ export interface ReadQueryResult {
   rows: Array<Row>;
 }
 
+export interface ReceiptResult {
```
joewagner, line 76: This is mainly for the sake of code readability since the response of an http request cannot be type checked via typescript.

src/lib/connector.ts
```diff
@@ -5,6 +5,7 @@ import { createToken } from "./token.js";
 import { query } from "./query.js";
 import { create } from "./create.js";
 import { hash } from "./hash.js";
+import { receipt } from "./tableland-calls.js";
```
joewagner, line 8: I tried to keep the function name short, but if anyone hash ideas lets discuss!

src/lib/create.ts
```diff
@@ -12,18 +12,16 @@ export async function create(
   this: Connection,
   query: string
 ): Promise<ContractReceipt> {
-  // TODO: depending on the outcome of some discussions this check to see if address is in
```
joewagner, line 15: As discussed internally we are no longer "gating" in this way.  There is a pending discussion of "attributing", but if that happend it will be a parameter in the SC call

src/lib/create.ts
```diff
@@ -12,18 +12,16 @@ export async function create(
...
   // This "dryrun" is done to validate that the query statement is considered valid.
   // We check this before minting the token, so the caller won't succeed at minting a token
   // then fail to create the table on the Tableland network
   await tablelandCalls.hash.call(this, query);
 
-  return await registerTable.call(this, query);
+  const txn = await registerTable.call(this, query);
 
   // TODO: we can potentially listen to Execution Tracker here and wait to return
+  //       until we have the receipt, but I'm sure if that makes sense.
```
joewagner, line 23: There was a longer internal discussion of waiting to return from this function until the Validator confirms that the transaction was processed.  I left the code to do that commented out so others can potentially investigate

src/lib/list.ts
```diff
@@ -4,7 +4,12 @@ export async function list(this: Connection): Promise<TableMetadata[]> {
   const address = await this.signer.getAddress();
 
   const resp: TableMetadata[] = await fetch(
-    `${this.host}/tables/controller/${address}`
+    `${this.host}/tables/controller/${address}`,
```
joewagner, line 7: In the v2 branch of the Valdator this bearer token is required.  See #43 for more detail

src/lib/tableland-calls.ts
```diff
@@ -65,7 +70,7 @@ async function GeneralizedRPC(
   const param: RpcRequestParam = {
     controller: address,
     create_statement: params.createStatement,
-    description: params.description,
```
joewagner, line 68: description is no longer supported in v2

src/lib/util.ts
```diff
@@ -14,6 +14,8 @@ const polyfills = {
 export const btoa = globalThis.btoa ?? polyfills.btoa;
 
 export const contractAddresses: { [index: string]: string } = {
+  local: "0x5fbdb2315678afecb367f032d93f642f64180aa3", // If building locally you can put your contract address here
```
joewagner, line 17: I added some more potential contracts and networks.  As we add deploy more contract+validator+gateways we can keep adding to this

carsonfarmer: It is also possible to _supply_ this from user land, right?

test/connect.test.ts
```diff
@@ -72,7 +72,7 @@ describe("connect function", function () {
         } as unknown as Signer // convince type checks into letting us mock the signer
       });
     } as ConnectionOptions).rejects.toThrow(
-      "Only Ethereum Rinkeby network is currently supported. Switch your wallet connection and reconnect."
+      "Only Ethereum Rinkeby, Optimism Kovan, and Local Hardhat networks are currently supported. Switch your wallet connection and reconnect."
```
carsonfarmer, line 75: Can we have this be generated from the phrases in the contractAddresses const, so we aren't hard "coding" this sentence? Maybe we can just create a separate issue for that, rather than holding up a PR with something as simple as this?

carsonfarmer (approved):
LGTM @joewagner. One comment about using the contractAddresses const to generate the error message, but nothing worth holding up this PR with!

src/lib/util.ts
joewagner: Yea, it will be after #45 
My thinking was that if we get to a place where running a fully local env is possible, having this will stream line things

PR #45, opened 2022-05-06
Allow specifying the contract as a connection option

joewagner:
This PR will expose an option to set the address of the contract you want to connect to.  This is being done to enable easier development and testing of the registry Smart Contract.

With this change the connection options would become:
 - token: Optionally pass in a Bearer token.  This can be used if the caller doesn't want the SDK to generate a SIWE token.  Potential uses include apps that save the user's SIWE token so they don't have to prompt the user to sign a new token on eevery page refresh
 - signer: Optionally pass in a Signer object.  This can be used if the caller wants to use a signer other than the browser default.  This is useful in the node.js enviroment, or really anywhere except within a browser that has a built-in wallet.
 - host:  This is the http url of the validator that the caller wants to connect to.  This is optional and defaults to https://testnet.tableland.network.  If host is not specified then network must be equal to 'testnet'
 - network: This is the name of the network, and as of this pr can be one of local, optimism-kovan-staging, staging, or testnet.  This name is mapped to the address of a contract that has been deployed and is being watched by a Tableland network Validator on that chain.  Note: 'local' is a locally running hardhat node, which requires that the caller has already got that running.
 - contract:  This PR will add the optional contract option.  If specified the SDKs direct calls to the SC will go to this address.  This is potentially useful for testing and local development.  Note: the abi of the contract must match the abi of the version of eth-tableland was used for the build of the SDK

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ const connection = await connect({
  - signer: Optionally pass in a Signer object. This can be used if the caller wants to use a signer other than the browser default. This is useful in the node.js enviroment, or really anywhere except within a browser that has a built-in wallet.
  - host: This is the http url of the validator that the caller wants to connect to. This is optional and defaults to https://testnet.tableland.network. If host is not specified then network must be equal to 'testnet'
  - network: This is the name of the network, and as of this pr can be one of local, optimism-kovan-staging, staging, or testnet. This name is mapped to the address of a contract that has been deployed and is being watched by a Tableland network Validator on that chain. Note: 'local' is a locally running hardhat node, which requires that the caller has already got that running. This is useful for contributors.
- - contract: This PR will add the optional contract option. If specified the SDKs direct calls to the SC will go to this address. This is potentially useful for testing and local development. Note: the abi of the contract must match the abi of the version of eth-tableland was used for the build of the SDK. This is useful for contributors.
+ - contract: Optionally pass in the address of a contract for making direct calls. If specified, the SDK's direct calls to the smart contract will go to _this_ address (rather than the default contract). This is potentially useful for testing and local development. Note: the ABI of the contract must match the ABI of the version of `eth-tableland` that was used to build the SDK. This is useful for contributors.
 
 ## Creating Tables
 
--- a/README.md
+++ b/README.md
@@ -84,6 +84,14 @@ const connection = await connect({
 });
 ```
 
+### Connection options
+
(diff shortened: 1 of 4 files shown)


src/lib/eth-calls.ts
```diff
@@ -12,12 +12,9 @@ async function registerTable(
   const address = await signer.getAddress();
   /* eslint-disable-next-line camelcase */
 
-  const network = this.network;
+  const contractAddress = this.contract ?? contractAddresses[this.network];
```
joewagner, line 15: The meat of this PR is here.  As a developer of the SDK or the SC you don't need to add the SC address to the [util.ts](https://github.com/tablelandnetwork/js-tableland/blob/joe/contract-op/src/lib/util.ts) file anymore.  You can pass it in a connection option

sanderpick:
🤩🤩

carsonfarmer (changes requested):
This looks great @joewagner. The preamble in this PR was extremely useful. The only change request I have is that you actually include this information in the doc string for the lib. That would be really nice to have, and would allow callers to decide exactly which approach they want/need (plus, docs = good).

README.md
```diff
@@ -84,6 +84,14 @@ const connection = await connect({
...
+ - host: This is the http url of the validator that the caller wants to connect to. This is optional and defaults to https://testnet.tableland.network. If host is not specified then network must be equal to 'testnet'
+ - network: This is the name of the network, and as of this pr can be one of local, optimism-kovan-staging, staging, or testnet. This name is mapped to the address of a contract that has been deployed and is being watched by a Tableland network Validator on that chain. Note: 'local' is a locally running hardhat node, which requires that the caller has already got that running. This is useful for contributors.
+ - contract: This PR will add the optional contract option. If specified the SDKs direct calls to the SC will go to this address. This is potentially useful for testing and local development. Note: the abi of the contract must match the abi of the version of eth-tableland was used for the build of the SDK. This is useful for contributors.
```
carsonfarmer, line 93:
```suggestion
 - contract: Optionally pass in the address of a contract for making direct calls. If specified, the SDK's direct calls to the smart contract will go to _this_ address (rather than the default contract). This is potentially useful for testing and local development. Note: the ABI of the contract must match the ABI of the version of `eth-tableland` that was used to build the SDK. This is useful for contributors.
```

carsonfarmer (approved):

PR #50, opened 2022-05-16
Fix regression in create method after custom contract option

joewagner:
When the custom contract address option was added there was a regression in the create method

### DIFF
--- a/src/lib/eth-calls.ts
+++ b/src/lib/eth-calls.ts
@@ -12,10 +12,12 @@ async function registerTable(
   const address = await signer.getAddress();
   /* eslint-disable-next-line camelcase */
 
-  const contractAddress = this.contract ?? contractAddresses[this.network];
+  const contractAddress = this.contract || contractAddresses[this.network];
+
+  if (!contractAddress)
+    throw new Error(`no contract found for ${this.network} network`);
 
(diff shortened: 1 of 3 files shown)


src/lib/eth-calls.ts
```diff
@@ -12,10 +12,12 @@ async function registerTable(
   const address = await signer.getAddress();
   /* eslint-disable-next-line camelcase */
 
-  const contractAddress = this.contract ?? contractAddresses[this.network];
+  const contractAddress = this.contract || contractAddresses[this.network];
```
joewagner, line 15: The issue was using the `??` which resulted in a cryptic error message saying 'network does not support ENS'. The value of `this.contract` was potentially an empty string which results in `contractAddress` being an empty string, which in turn is interpreted by ethersproject as an ENS domain.

test/connect.test.ts
```diff
@@ -34,6 +40,20 @@ describe("connect function", function () {
     await expect(typeof connection.create).toBe("function");
   });
 
+  test("allows specifying connection network", async function () {
```
joewagner, line 43: Added a test for good measure

awmuncy (approved):
Looks good!

PR #51, opened 2022-05-23
Updates to match changes to RPC endpoint

joewagner:
These updates include changes to the create api, and the splitting of the query method into a read, and a write method.

The thinking behind the changes to the create api is discussed here: https://www.notion.so/textile/Align-the-intention-more-directly-with-the-outcome-when-creating-tables-a82b38c7606942e48f9b3f6a033f516f#e1461d1d49574c92ab00d9d3e2674840

The thinking behind the splitting of `query` into `read` and `write` is discussed here: https://www.notion.so/textile/Allow-unauthenticated-reads-over-RPC-and-split-RunSQL-API-43f3890ec4d94755a319a27928e7efdd

One of the bigger related Validator PRs is here https://github.com/textileio/go-tableland/pull/114

A change to the Registry contract is needed to enable the new runSQL flow.  A potential for that change can be found here: https://github.com/tablelandnetwork/eth-tableland/pull/15

There are quite a few moving pieces here, and I didn't feel like it was straight forward to see if they are all working together.  Because of this I created a was to run a few "end-to-end", or "smoke-screen", tests here: https://github.com/textileio/local-tableland

### DIFF
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -69,6 +69,10 @@ export type Column = Array<ColumnDescriptor>;
 
 export type Row = Array<string | number | boolean>;
 
+export interface WriteQueryResult {
+  data: null;
+}
+
 export interface ReadQueryResult {
   columns: Array<Column>;
   rows: Array<Row>;
@@ -131,9 +135,8 @@ export interface Connection {
      **/
     prefix?: string
   ) => Promise<ContractReceipt>;
-  query: (query: string) => Promise<null | ReadQueryResult>; // TODO
-  // read: (query: string) => Promise<null | ReadQueryResult>;
-  // write: (query: string) => Promise<null | {data: null}>;
+  read: (query: string) => Promise<null | ReadQueryResult>;
+  write: (query: string) => Promise<null | WriteQueryResult>;
   hash: (query: string) => Promise<StructureHashReceipt>;
   receipt: (txnHash: string) => Promise<ReceiptResult>;
 }
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -39,11 +39,11 @@ export type KeyVal<T = any> = [string, T];
 
 export type Column = Array<{ name: string }>;
 
-export type Row = Array<string | number | boolean>;
-
-export interface ReadQueryResult {
+export interface ReadQueryResult<
+  Row extends Array<string | number | boolean> = Array<any>
+> {
   columns: Array<Column>;
-  rows: Array<Row>;
+  rows: Row;
 }
 
 export interface WriteQueryResult {
@@ -106,5 +106,5 @@ export interface Connection {
   read: (query: string) => Promise<ReadQueryResult>;
   write: (query: string) => Promise<WriteQueryResult>;
   hash: (query: string) => Promise<StructureHashReceipt>;
-  receipt: (txnHash: string) => Promise<ReceiptResult | void>;
+  receipt: (txnHash: string) => Promise<ReceiptResult | undefined>;
 }
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -1,14 +1,11 @@
 import { BigNumber, ContractReceipt, Signer } from "ethers";
 
 export interface TableMetadata {
-  id: string;
+  controller: string;
   /* eslint-disable-next-line camelcase */
-  created_at?: string;
-  description?: string;
-  tablename?: string;
-  name?: string;
-  controller?: string;
-  structure?: string;
+  created_at: string;
+  name: string;
+  structure: string;
 }
 
 export interface Token {
@@ -24,30 +21,12 @@ export interface ConnectionOptions {
 }
 
 export interface RpcParams {
-  controller?: boolean;
-  createStatement?: string;
-  description?: string;
-  dryrun?: boolean;
-  statement?: string;
-  tableId?: string;
-  txnHash?: string;
-}
-
-export interface RpcRequestParam {
   controller?: string;
   /* eslint-disable-next-line camelcase */
   create_statement?: string;
-  description?: string;
-  dryrun?: boolean;
-  id?: string;
   statement?: string;
   /* eslint-disable-next-line camelcase */
   txn_hash?: string;
-}
-
-export interface ConnectionReceipt {
-  jwsToken: Token;
-  ethAccounts: Array<string>;
 }
 
 export interface SupportedNetwork {
@@ -56,26 +35,19 @@ export interface SupportedNetwork {
   chainId: number;
 }
 
-/**
- * ColumnDescriptor gives metadata about a colum (name, type)
- */
-export interface ColumnDescriptor {
-  name: string;
-}
-
 export type KeyVal<T = any> = [string, T];
 
-export type Column = Array<ColumnDescriptor>;
+export type Column = Array<{ name: string }>;
 
 export type Row = Array<string | number | boolean>;
-
-export interface WriteQueryResult {
-  data: null;
-}
 
 export interface ReadQueryResult {
   columns: Array<Column>;
   rows: Array<Row>;
+}
+
+export interface WriteQueryResult {
+  hash: string;
 }
 
 export interface ReceiptResult {
@@ -118,16 +90,12 @@ export interface Connection {
   contract: string;
   list: () => Promise<TableMetadata[]>;
   create: (
-    /** The ChainId of the chain this table should be created on.
-     *  MUST match the chain that the SDK connected to
-     **/
-    chainId: number,
     /** The schema that defines the columns and constraints of the table,
      *  e.g.
-     * `id int NOT NULL,
-     *  name char(50) NOT NULL,
-     *  favorite_food char(50),
-     *  PRIMARY KEY (id)`
+     `id int NOT NULL,
+      name char(50) NOT NULL,
+      favorite_food char(50),
+      PRIMARY KEY (id)`
      */
     schema: string,
     /** an optional prefix to the tablename that will be assigned to this table.
@@ -135,8 +103,8 @@ export interface Connection {
      **/
     prefix?: string
   ) => Promise<ContractReceipt>;
-  read: (query: string) => Promise<null | ReadQueryResult>;
-  write: (query: string) => Promise<null | WriteQueryResult>;
+  read: (query: string) => Promise<ReadQueryResult>;
+  write: (query: string) => Promise<WriteQueryResult>;
   hash: (query: string) => Promise<StructureHashReceipt>;
-  receipt: (txnHash: string) => Promise<ReceiptResult>;
+  receipt: (txnHash: string) => Promise<ReceiptResult | void>;
 }
--- a/src/interfaces.ts
+++ b/src/interfaces.ts
@@ -84,15 +84,6 @@ export interface ReceiptResult {
   error?: string;
 }
 
-export interface CreateTableOptions {
-  /** A human readable description of the nature and purpoe of the table */
-  description?: string;
-  /** If your table was minted, but never created on tableland, use this param to create it. */
-  id?: string;
-  /** do a dry run of create to see if the create statement is valid without creating the table */
-  dryrun?: boolean;
-}
-
 export interface CreateTableReceipt {
   name: string;
   structureHash: string;
@@ -123,10 +114,26 @@ export interface Connection {
   contract: string;
   list: () => Promise<TableMetadata[]>;
   create: (
-    query: string,
-    options?: CreateTableOptions
+    /** The ChainId of the chain this table should be created on.
+     *  MUST match the chain that the SDK connected to
+     **/
+    chainId: number,
+    /** The schema that defines the columns and constraints of the table,
+     *  e.g.
+     * `id int NOT NULL,
+     *  name char(50) NOT NULL,
+     *  favorite_food char(50),
+     *  PRIMARY KEY (id)`
+     */
+    schema: string,
+    /** an optional prefix to the tablename that will be assigned to this table.
+     *  If supplied, it must conform to the rules of SQL table names
+     **/
+    prefix?: string
   ) => Promise<ContractReceipt>;
-  query: (query: string) => Promise<null | ReadQueryResult>;
+  query: (query: string) => Promise<null | ReadQueryResult>; // TODO
+  // read: (query: string) => Promise<null | ReadQueryResult>;
+  // write: (query: string) => Promise<null | {data: null}>;
   hash: (query: string) => Promise<StructureHashReceipt>;
   receipt: (txnHash: string) => Promise<ReceiptResult>;
 }
--- a/src/lib/tableland-calls.ts
+++ b/src/lib/tableland-calls.ts
@@ -105,7 +105,7 @@ async function write(
 async function receipt(
   this: Connection,
   txnHash: string
-): Promise<ReceiptResult | void> {
+): Promise<ReceiptResult | undefined> {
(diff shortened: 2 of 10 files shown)


src/interfaces.ts
```diff
@@ -123,10 +118,25 @@ export interface Connection {
...
   list: () => Promise<TableMetadata[]>;
   create: (
-    query: string,
-    options?: CreateTableOptions
+    /** The ChainId of the chain this table should be created on.
+     *  MUST match the chain that the SDK connected to
+     **/
+    chainId: number,
+    /** The schema that defines the columns and constraints of the table,
+     *  e.g.
```
carsonfarmer, line 126: The example here isn't a string. I think you'll want to use real JS formatting so folks could (in theory) copy paste this code, because that's what they'll end up doing.

src/interfaces.ts
```diff
@@ -69,6 +69,10 @@ export type Column = Array<ColumnDescriptor>;
 
 export type Row = Array<string | number | boolean>;
 
+export interface WriteQueryResult {
+  data: null;
```
carsonfarmer, line 73: If this is always going to be `null`, then we just shouldn't return anything in any externally facing APIs to user-land. That just ends up being confusing. Not sure if this _is_ the case, I'll see as I read on. But if we're _ever_ returning this to user land, we should just return `void` in typescript speak, which is effectively `undefined`.

src/interfaces.ts
```diff
@@ -123,10 +118,25 @@ export interface Connection {
...
+     */
+    schema: string,
+    /** an optional prefix to the tablename that will be assigned to this table.
+     *  If supplied, it must conform to the rules of SQL table names
+     **/
+    prefix?: string
   ) => Promise<ContractReceipt>;
-  query: (query: string) => Promise<null | ReadQueryResult>;
+  read: (query: string) => Promise<null | ReadQueryResult>;
+  write: (query: string) => Promise<null | WriteQueryResult>;
```
carsonfarmer, line 139: If this is an external API, I'd rather just see `Promise<void>` as the return type here. Or something else useful if not that. But or-ing here is just an odd API response type.

src/interfaces.ts
```diff
@@ -123,10 +118,25 @@ export interface Connection {
...
+     *  PRIMARY KEY (id)`
+     */
+    schema: string,
+    /** an optional prefix to the tablename that will be assigned to this table.
+     *  If supplied, it must conform to the rules of SQL table names
+     **/
+    prefix?: string
   ) => Promise<ContractReceipt>;
-  query: (query: string) => Promise<null | ReadQueryResult>;
+  read: (query: string) => Promise<null | ReadQueryResult>;
```
carsonfarmer, line 138: Why would null show up? I'd rather not have `null` in our public APIs, as this is frowned upon in typescript. It would be better to always be a ReadQueryResult or error if something goes wrong.

src/lib/tableland-calls.ts
```diff
@@ -120,6 +121,19 @@ async function query(
...
+  this: Connection,
+  query: string
+): Promise<WriteQueryResult | null> {
+  const message = await GeneralizedRPC.call(this, "relayWriteQuery", {
+    statement: query,
+  });
+  const response = await SendCall.call(this, message);
+  const json = await sendResponse(response);
+
+  return json as WriteQueryResult;
```
carsonfarmer, line 134: Here, I'd just `return` and have the response be a `Promise<void>`. Line 132 could also just be `await sendResponse(response);`

test/connect.test.ts
```diff
@@ -47,8 +48,7 @@ describe("connect function", function () {
     fetch.mockResponseOnce(FetchCreateDryRunSuccess);
     fetch.mockResponseOnce(FetchCreateTableOnTablelandSuccess);
 
-    const createStatement = "CREATE TABLE hello (id int primary key, val text);";
-    await connection.create(createStatement);
+    await connection.create(31337, "id int primary key, val text", "hello");
```
carsonfarmer, line 51: ah, so nice!

carsonfarmer (changes requested):
Just a few minor changes that I'd like to see @joewagner. Since we're breaking our API anyway, we might as well make a nice one we'll be really happy with. `Promise<void>` is nice for APIs that are just an async op with only "side effects", whereas, we should avoid union response types if one of the types is null or undefined, and instead error out when we have an empty response (unless this is an expected thing).

carsonfarmer:
_Personally_ I think we could go with a generic on the read query function, so that way, callers can specify the type they are expecting in the response. Lemme know if you wanna discuss this, because I think it would increase usability of the API a great deal.

test/connect.test.ts
joewagner: Absolutey!  The new arity is much nicer

joewagner:
> _Personally_ I think we could go with a generic on the read query function, so that way, callers can specify the type they are expecting in the response. Lemme know if you wanna discuss this, because I think it would increase usability of the API a great deal.

Great points, thanks for all the feedback.  I'll try to clean up and make some changes, then ping you again and maybe discuss.

src/lib/tableland-calls.ts
carsonfarmer: If the returned value of a write _does_ provide useful information about the right (rows written or something?), then we _should_ return this, because it provides useful information. Otherwise, don't.

src/interfaces.ts
```diff
@@ -123,10 +118,25 @@ export interface Connection {
   contract: string;
   list: () => Promise<TableMetadata[]>;
   create: (
-    query: string,
-    options?: CreateTableOptions
+    /** The ChainId of the chain this table should be created on.
+     *  MUST match the chain that the SDK connected to
+     **/
+    chainId: number,
```
carsonfarmer, line 124: I see that this is required. Could we don't default to whatever chained the sdk is connected to in the public APIs? This information can be pulled from the ethers connection.

src/lib/tableland-calls.ts
joewagner: The validator RPC response is `{data: null}` which does seem like we might as well return void.  
I recognize I'm throwing a jellyfish here, but it seems related and worth mentioning.  Now that the validator is acting as a relay to the Smart Contract and not actually doing the write until the SC emits the event, this function is returning before we have full confidence the write is finished.

src/interfaces.ts
joewagner: I was thinking the same thing, but the TPD docs mention making this required.  Maybe I was reading the intent wrong?  We can absolutely remove this argument all together.  The only reasons to include it afaik is so the caller is forced to be explicit, and maybe in the future calls to create could be direct to the SC without any validator connection.
Honestly I'd be happy removing it.
here is the Notion doc that I was referencing: https://www.notion.so/textile/Align-the-intention-more-directly-with-the-outcome-when-creating-tables-a82b38c7606942e48f9b3f6a033f516f#e1461d1d49574c92ab00d9d3e2674840

carsonfarmer: I think in the TPD doc, the _explicit_ talk was more about the APIs on the validators (though I see it _is_ included in the mention of the SDK). In any case, the SDKs should paper over this kind of thing. Even with calls direct to the smart contract, the SDK could provide a default.

carsonfarmer: If you search through the resolved comments there, we mention something about the SDK letting the user avoid adding chain id and other things that are always fixed.

src/lib/tableland-calls.ts
carsonfarmer: Ah gotcha, ok. Let's just document it for now then (on the method itself as doc strings), and then you can create a tension for it when you have time 👍 .

carsonfarmer: Example: Note: This method returns right away, once the write request has been sent to a validator for writing to the Tableland smart contract. However, the write is not confirmed until a validator has picked up the write event from the smart contract, and digested the event locally.

joewagner: Sorry this thread is getting long...
I was in process of making these updates, and I see that the validator now returns the transaction hash.  I feel like it would make sense for the SDK to return that as either a string or an object with a single property. Thoughts?

src/interfaces.ts
```diff
@@ -123,10 +90,21 @@ export interface Connection {
   contract: string;
   list: () => Promise<TableMetadata[]>;
   create: (
-    query: string,
-    options?: CreateTableOptions
+    /** The schema that defines the columns and constraints of the table,
```
joewagner, line 93: chainId is gone, it's being sniffed out of signer.provider

src/interfaces.ts
```diff
@@ -1,14 +1,11 @@
 import { BigNumber, ContractReceipt, Signer } from "ethers";
 
 export interface TableMetadata {
```
joewagner, line 3: Lots of clean up of interfaces that are no longer needed and changing of the props to match what the validator sends

src/interfaces.ts
```diff
@@ -123,10 +90,21 @@ export interface Connection {
...
+     *  If supplied, it must conform to the rules of SQL table names
+     **/
+    prefix?: string
   ) => Promise<ContractReceipt>;
-  query: (query: string) => Promise<null | ReadQueryResult>;
+  read: (query: string) => Promise<ReadQueryResult>;
+  write: (query: string) => Promise<WriteQueryResult>;
   hash: (query: string) => Promise<StructureHashReceipt>;
-  receipt: (txnHash: string) => Promise<ReceiptResult>;
+  receipt: (txnHash: string) => Promise<ReceiptResult | void>;
```
joewagner, line 109: If the transaction hasn't been digested return undefined

src/lib/tableland-calls.ts
```diff
@@ -1,55 +1,44 @@
...
       "Content-Type": "application/json",
       Authorization: `Bearer ${this.token.token}`,
     },
     body: JSON.stringify(rpcBody),
   });
+
+  return parseResponse(res);
 }
 
 // parse the rpc response and throw if any of the different types of errors occur
```
joewagner, line 27: I refactored this so that the different structures returned by the RPC calls are handled in the method's body.  This way we don't have to do stuff like `return json.some.key as SomeType`

src/lib/tableland-calls.ts
```diff
@@ -83,54 +66,53 @@ async function GeneralizedRPC(
...
+  return camelCaseKeys(json.result);
+}
 
-  const response = await SendCall.call(this, message);
-  const json = await sendResponse(response);
+async function read(this: Connection, query: string): Promise<ReadQueryResult> {
+  const message = await GeneralizedRPC.call(this, "runReadQuery", {
+    statement: query,
+  });
+  const json = await SendCall.call(this, message);
```
joewagner, line 86: no longer using the as operator to fix type assertion errors

src/lib/tableland-calls.ts
```diff
@@ -83,54 +66,53 @@ async function GeneralizedRPC(
...
+    txn_hash: txnHash,
   });
-  const response = await SendCall.call(this, message);
-  const json = await sendResponse(response);
+  const json = await SendCall.call(this, message);
 
-  return json as ReceiptResult;
+  if (json.result.receipt) return camelCaseKeys(json.result.receipt);
+  // if the transaction hasn't been digested return undefined
 }
```
joewagner, line 117: Not sure why we were exporting `TableMetadata`, but we don't need to do that anymore

src/interfaces.ts
carsonfarmer: Nice, this is much cleaner.

src/interfaces.ts
carsonfarmer: Perfect. This looks great.

src/interfaces.ts
```diff
@@ -24,48 +21,23 @@ export interface ConnectionOptions {
...
- * ColumnDescriptor gives metadata about a colum (name, type)
- */
-export interface ColumnDescriptor {
-  name: string;
-}
-
 export type KeyVal<T = any> = [string, T];
 
-export type Column = Array<ColumnDescriptor>;
+export type Column = Array<{ name: string }>;
```
carsonfarmer, line 40: So much red here, I love it!

src/interfaces.ts
```diff
@@ -74,6 +46,10 @@ export interface ReadQueryResult {
   rows: Array<Row>;
 }
 
+export interface WriteQueryResult {
+  hash: string;
```
carsonfarmer, line 50: Cool, this is useful.

src/interfaces.ts
carsonfarmer: Void should be reserved for functions that don't return _anything_, rather than sometimes return undefined: https://www.typescripttutorial.net/typescript-tutorial/typescript-void-type/

```suggestion
  receipt: (txnHash: string) => Promise<ReceiptResult | undefined>;
```

src/lib/tableland-calls.ts
carsonfarmer: Yeh nice.

src/lib/tableland-calls.ts
```diff
@@ -83,54 +66,53 @@ async function GeneralizedRPC(
...
 
-  return json as ReadQueryResult;
+  return camelCaseKeys(json.result.tx);
 }
 
 async function receipt(
   this: Connection,
   txnHash: string
-): Promise<ReceiptResult> {
+): Promise<ReceiptResult | void> {
```
carsonfarmer, line 108: undefined here rather than void

test/connect.test.ts
```diff
@@ -47,8 +48,7 @@ describe("connect function", function () {
     fetch.mockResponseOnce(FetchCreateDryRunSuccess);
     fetch.mockResponseOnce(FetchCreateTableOnTablelandSuccess);
 
-    const createStatement = "CREATE TABLE hello (id int primary key, val text);";
-    await connection.create(createStatement);
+    await connection.create("id int primary key, val text", "hello");
```
carsonfarmer, line 51: thing of beauty

src/interfaces.ts
```diff
@@ -24,48 +21,23 @@ export interface ConnectionOptions {
...
-export interface ColumnDescriptor {
-  name: string;
-}
-
 export type KeyVal<T = any> = [string, T];
 
-export type Column = Array<ColumnDescriptor>;
+export type Column = Array<{ name: string }>;
 
 export type Row = Array<string | number | boolean>;
```
carsonfarmer, line 43: `ReadQueryInterface` could be made generic in rows. Such that:

```
export interface ReadQueryResult<Row extends Array<string | number | boolean> = Array<any>> {
  columns: Array<Column>;
   rows: Row;
 }
```

And we delete the `Row` type above (because it is no longer needed).

Does that seem useful?

carsonfarmer (approved):
Lots of awesome changes here. A few comments to help clean things up, but nothing particularly major.

src/lib/tableland-calls.ts
carsonfarmer:
```suggestion
): Promise<ReceiptResult | undefined> {
```

src/interfaces.ts
carsonfarmer: Just a suggestion to make read generic, so you can do things like:

```
type MyType = [number, string, number]
const response = await sdk.read<MyType>("blah blah")
```

where response contains an array of rows that are typed as MyType.

joewagner: Love this!

joewagner:
@carsonfarmer Just pushed again, I think I got everything now.  If you wanna give the the 👍  I'll merge

Issue #22, opened 2022-02-20
Checks on signer provider are currently too restrictive.


carsonfarmer:
This makes it difficult to use with none browser-based providers, or signers that don't require a remote provider.

joewagner:
fixed via https://github.com/tablelandnetwork/js-tableland/commit/92ac487bdb05cbad2dc5eea92a0cda7875dbddab

PR #53, opened 2022-05-27
Update SDK to match go-tableland v2

joewagner:
Once https://github.com/textileio/go-tableland/tree/v2 is merged into main, we can begin do the same merge in this repo. 

The v2 branch will need to be updated to include the contract addresses where eth-tableland was deployed.  These updates will be made in the src/lib/util.ts file.  The @tableland/eth dependency should be updated to be “0.1.0” matching the package version which will be published once eth-tableland is ready.
Once this is done the v2 branch should be merged into the main branch and all build, lint, and test should work.  A new version, 0.1.0, of this repo should be published to npm with the latest tag.

Steps to publish to npm are documented here: https://www.notion.so/textile/Publishing-to-NPM-d6051eaab7df4673949cd3dffc322806


carsonfarmer:
Hey @joewagner, I just pushed the updated eth-tableland release. The checks are passing. I think we can merge and publish this one.

src/lib/util.ts
```diff
@@ -14,23 +14,36 @@ const polyfills = {
 export const btoa = globalThis.btoa ?? polyfills.btoa;
 
 export const contractAddresses: { [index: string]: string } = {
```
sanderpick, line 16: to be most consistent / clear, i think we should just match the naming pattern here: https://github.com/tablelandnetwork/eth-tableland/blob/main/hardhat.config.ts#L40

otherwise, "testnet" and "staging" are ambiguous

sanderpick (commented):
so much in here! one comment regarding your network naming question

src/lib/util.ts
sanderpick: I believe we can also leave out "optimism-kovan-staging" since that should be for internal use only. and since you changed the user API to be flexible, it would be easy for use to manually use this in our internal testing

carsonfarmer: Agree that naming conventions should be consistent. But actually, we probably should have gone with these "standards": https://github.com/ethers-io/ethers.js/blob/master/packages/networks/src.ts/index.ts#L132. Many other projects use these apparently. For now, I'm happy just creating a mapping from what we're using to these internally (for the cli, these names are required).

carsonfarmer: The above link references https://chainlist.org

sanderpick: there's more information to capture though than what's in those names. ie, each tableland-to-chain pair, where the chain can be used twice, as we are doing with optimism kovan (and others in the future). for the CLI, maybe you can separate those, eg `--goerli --testnet` points to goerli on tableland testnet, `--goerli --staging` would point to goerli on tableland staging

carsonfarmer: See my recent pushes. As far as the cli goes, its actually even more complicated than the above would suggest. We need to specify the contract address, _and_ the provider API to connect with. I can find some nice mappings to use, but not tonight :P

src/lib/connector.ts
```diff
@@ -69,13 +49,22 @@ export async function connect(options: ConnectionOptions): Promise<Connection> {
 
   if (
     !providerNetwork?.name ||
-    !SUPPORTED_NETWORKS.find((net) => net.key === providerNetwork.name)
+    !SUPPORTED_NETWORKS.find((net) => {
```
carsonfarmer, line 52: Why is supported networks not just a map/object?

src/lib/connector.ts
```diff
@@ -1,62 +1,42 @@
...
+    chainId: chainId,
+    expirationTime: exp,
+    uri: globalThis.document?.location.origin,
+    version: "1",
+    statement: "Official Tableland SDK",
+  });
 }
 
 export async function connect(options: ConnectionOptions): Promise<Connection> {
   const network = options.network ?? "testnet";
```
carsonfarmer, line 37: is this right?

src/interfaces.ts
```diff
@@ -20,77 +17,60 @@ export interface ConnectionOptions {
...
-  name: string;
+  chainId: number;
 }
 
 export type KeyVal<T = any> = [string, T];
 
-export type Column = Array<ColumnDescriptor>;
-
-export type Row = Array<string | number | boolean>;
+export type Columns = Array<{ name: string }>;
```
carsonfarmer, line 41: Fixing this to match actual response. Looks like there was a "double" wrapping in an Array.

src/interfaces.ts
```diff
@@ -20,77 +17,60 @@ export interface ConnectionOptions {
...
 
-export interface ReadQueryResult {
-  columns: Array<Column>;
-  rows: Array<Row>;
+export interface ReadQueryResult<
+  Rows extends Array<string | number | boolean> = Array<any>
+> {
+  columns: Columns;
+  rows: Rows;
```
carsonfarmer, line 47: Same deal here.

src/interfaces.ts
```diff
@@ -20,77 +17,60 @@ export interface ConnectionOptions {
...
 
-export interface CreateTableReceipt {
-  name: string;
-  structureHash: string;
-  description?: string;
+export interface ReceiptResult {
+  /* eslint-disable-next-line camelcase */
+  chain_id: number;
+  /* eslint-disable-next-line camelcase */
+  txn_hash: string;
```
carsonfarmer, line 58: Is this right? In practice it appears to be CamelCase on the outgoing. Is this just due to your camelCase code @joewagner ?

src/interfaces.ts
```diff
@@ -20,77 +17,60 @@ export interface ConnectionOptions {
...
 }
 
 export interface StructureHashReceipt {
   structureHash: string;
 }
 
-export interface TableRegistrationReceipt {
-  receipt: ContractReceipt;
-  tableId: BigNumber;
+export interface CreateTableReceipt {
```
carsonfarmer, line 68: Check me on this. I've created this as an output to create. Check out the new function sig and tests.

src/interfaces.ts
```diff
@@ -104,11 +84,24 @@ export interface Connection {
...
+      name char(50) NOT NULL,
+      favorite_food char(50),
+      PRIMARY KEY (id)`
+     */
+    schema: string,
+    /** an optional prefix to the tablename that will be assigned to this table.
+     *  If supplied, it must conform to the rules of SQL table names
+     **/
+    prefix?: string
   ) => Promise<CreateTableReceipt>;
```
carsonfarmer, line 102: New return type here

src/lib/connector.ts
```diff
@@ -1,64 +1,44 @@
...
-  // WARN: This is a non-standard JWT
-  // Borrows ideas from: https://github.com/ethereum/EIPs/issues/1341
-  const iss = await signer.getAddress();
-  const network = await signer.provider?.getNetwork();
-  const chain = network?.chainId ?? "unknown";
-  let net = network?.name;
-  if (net?.startsWith("matic")) net = "poly";
-  else net = "eth";
-  const kid = `${net}:${chain}:${iss}`;
+export async function userCreatesToken(
```
carsonfarmer, line 20: Formatting fix, and also this is now accessible from outside the lib. It is useful for creating standalone tokens.

src/lib/connector.ts
```diff
@@ -1,64 +1,44 @@
...
 
 export async function connect(options: ConnectionOptions): Promise<Connection> {
-  const network = options.network ?? "testnet";
-  const host = options.host ?? "https://testnet.tableland.network";
+  const network = options.network ?? "goerli";
+  const host = options.host ?? "https://testnetv2.tableland.network";
+  const contract = options.contract ?? contractAddresses[network];
 
-  if (network !== "testnet" && !options.host) {
+  if (network !== "goerli" && !options.host) {
```
carsonfarmer, line 41: updating the defaults

src/lib/create.ts
```diff
@@ -1,40 +1,34 @@
...
+ * @param {string} query SQL create statement.
+ * @returns {string} The smart contract transaction receipt.
  */
 export async function create(
   this: Connection,
-  query: string,
-  options: CreateTableOptions = {}
+  schema: string,
+  prefix: string = ""
 ): Promise<CreateTableReceipt> {
```
carsonfarmer, line 15: Notice that, while this hasn't changed, the interface itself has.

src/lib/create.ts
```diff
@@ -1,40 +1,34 @@
...
 
-  let id = options.id;
-  if (!id) {
-    const { tableId } = await registerTable.call(this);
-    id = BigNumber.from(tableId).toString();
-  }
+  const txn = await registerTable.call(this, query);
 
-  return await tablelandCalls.create.call(this, query, id, options);
+  const [, event] = txn.events ?? [];
```
carsonfarmer, line 29: Since we're getting the transaction anyway, why not just extract the stuff from the event? The validator might not have picked it up yet, but that's ok... we have the transaction txnHash to use to check for status later.

src/lib/util.ts
carsonfarmer: Bumping this one again, just to make sure it looks like what y'all were expecting? In the CLI, essentially you have to specify via the "host" param if you are using staging or testnet, and it should pick the right one from that.

test/createTable.test.ts
```diff
@@ -12,9 +12,10 @@ describe("create method", function () {
   beforeAll(async function () {
     // reset in case another test file hasn't cleaned up
     fetch.resetMocks();
+    // const signer = ethers.providers.Web3Provider().getSigner();
     connection = await connect({
-      network: "testnet",
-      host: "https://testnet.tableland.network",
+      network: "goerli",
+      host: "https://testnetv2.tableland.network",
```
carsonfarmer, line 18: correct defaults

carsonfarmer (commented):
Hey @sanderpick and @joewagner can you both take another look at this one? There are lots of new changes from my testing with the CLI etc. This was my first hands-on work on the SDK in a while, so if you don't think some of the changes are useful, I'm open to alternatives.

joewagner:
> Hey @sanderpick and @joewagner can you both take another look at this one? There are lots of new changes from my testing with the CLI etc. This was my first hands-on work on the SDK in a while, so if you don't think some of the changes are useful, I'm open to alternatives.

Looks fantastic. I left a few comments, but nothing that needs changed.  🏁

src/lib/util.ts
sanderpick: taking a look

src/lib/connector.ts
```diff
@@ -1,64 +1,44 @@
...
     throw Error(
       "Please specify a host to connect to. (Example: https://env.tableland.network)"
```
sanderpick, line 43: the example might be confusing. i could see people trying to use a host of `https://env.tableland.network`

src/lib/util.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    return buffer.toString("base64");
+  },
+};
+
+export const btoa = globalThis.btoa ?? polyfills.btoa;
+
+export const contractAddresses: { [index: string]: string } = {
+  // If building locally you can put your contract address here or use the contract connection option
+  localhost: "0xe7f1725e7734ce288f8367e1bb143e90bb3f0512",
```
sanderpick, line 18: probably best to remove the placeholder string for localhost... could be empty string or some obvious placeholder, eg, "LOCAL_HARDHAT_CONTRACT". actually, could even come from env?

sanderpick (approved):
awesome stuff guys! i left a couple nits, but adding a +1

src/lib/util.ts
carsonfarmer: We don't want to assume an ENV here at all, because this will also run in a browser. But the CLI can assume this and fill it in.

src/lib/connector.ts
carsonfarmer: yeh

carsonfarmer: let's leave for now... but if we find we end up just iterating over it a lot, we might consider doing the object route.

src/lib/connector.ts
carsonfarmer: hmm, I guess it depends on the semantics of the network arg. is that the tableland network, or the chain network?

src/interfaces.ts
carsonfarmer: yes please!

src/interfaces.ts
carsonfarmer: No no camelCase is fine, and if this is just an internal interface, then this is also fine. But we should just make sure all our external interfaces are camelCase.

src/lib/connector.ts
carsonfarmer: So here's the question then, is network here mean to identify the tableland network, or the chain network. Since host already has the tableland network information in it, perhaps we use network to identify the chain network?

src/lib/connector.ts
carsonfarmer: updated

src/lib/util.ts
carsonfarmer: It is only deterministic in that it is a fresh address and deploying the contract is the first thing you do with that address. But that might not always be the case.

src/lib/connector.ts
joewagner: After our chat today it occurs to me that this might be related to #22 and the signer checks for provider as well.  
TLDR; I think what we have here works well, but thinking through all the connection params again in combination with #22 could be good.

Some more details:

The SDK needs the following to enable **ALL** the method calls:
 1. An http url for the validator/gateway that the caller wants to connect to Tableland with.  Note that for everything to work this validator is going to have to support the chain in number 5 below and share the smart contract address.
 2. A signer object to use for proving the user's identity to the Validator, EVM, etc...
 3. If the caller wants to call `create` a provider object, as defined by `ethers.js`, is needed.  This is used for any direct interaction with the EVM.  At the moment this is only the create method, but I can imagine other cases like directly setting or unsetting the controller.  If we stop doing the strict checks on signer this could be optional and then error if the create method is called on a connection that doesn't have a provider.
 4. The address of the Tableland Tables smart contract deployment (registry contract) that will be shared between the SDK caller and the Validator that the SDK caller specified above.  
 5. The chainId of the EVM where the shared contract is deployed

The combination of 1, 4, and 5 is what I've been thinking about as a Tableland Network.  I could see changing the connection params to enable something like `connect({network: 'goerli-testnet'})` where the validator/gateway url is set to `https://testnetv2.tableland.network/` the contract address is set to `0xsdasfdwhatever`, and the chainId is set to `5`.  If the __optional__ provider's getNetwork method exists and doesn't match the specified Tableland Network's requirement then we would have to error out.
Sorry that's is all a lot for a cr comment.  Maybe we scope this out in #22?

carsonfarmer: Naw this is super helpful! I'm going to finish this up tomorrow and we'll see what everyone thinks. I like this approach of only requiring a provider when explicitly needed.

PR #55, opened 2022-06-11
feat: new user ex for create + lazy sign and tokens

carsonfarmer:


src/lib/connection.ts
```diff
@@ -0,0 +1,73 @@
+import { BigNumber, Signer } from "ethers";
```
carsonfarmer, line 1: Just pulls this into one place

src/lib/connector.ts
```diff
@@ -1,97 +1,93 @@
...
-  if (network !== "goerli" && !options.host) {
-    throw Error(
-      "Please specify a host connection (e.g. https://testnetv2.tableland.network)"
-    );
+/**
+ * Create client connection with Tableland, EVM, and Gateway.
+ * @param options Options to control client connection.
+ * @returns Promise that resolves to a Connection object.
+ */
+export async function connect(options: ConnectOptions): Promise<Connection> {
```
carsonfarmer, line 37: Ok, so this makes it so that you can't connect without supplying a signer right away. It will then lazily check for a signer only when it needs it. This then supports a readonly client mode where you can make queries without ever signing anything.

src/lib/eth-calls.ts
```diff
@@ -1,21 +1,17 @@
...
-import { contractAddresses } from "./util.js";
+import { Connection } from "./connection.js";
+import { getSigner } from "./util.js";
 
 async function registerTable(
   this: Connection,
   query: string
 ): Promise<ContractReceipt> {
-  const signer = this.signer;
+  const signer = this.signer ?? (await getSigner());
```
carsonfarmer, line 11: lazily get signer if not provided. this will throw in node if a signer isn't provided

src/lib/tableland-calls.ts
```diff
@@ -1,24 +1,41 @@
...
+  result: T;
+}
 
-async function SendCall(this: Connection, rpcBody: Object) {
-  const res = await fetch(`${this.host}/rpc`, {
+async function SendCall(this: Connection, rpcBody: Object, token?: string) {
+  const headers: Record<string, string> = {
+    "Content-Type": "application/json",
+  };
+  if (token) {
```
carsonfarmer, line 31: we only sometimes need this

src/lib/tableland-calls.ts
```diff
@@ -36,44 +53,32 @@ async function parseResponse(res: any): Promise<any> {
...
-
 async function GeneralizedRPC(
   this: Connection,
   method: string,
   params: RpcParams = {}
 ) {
-  const signer = this.signer;
-  const address = await signer.getAddress();
-
-  const param: RpcParams = { controller: address, ...params };
```
carsonfarmer, line 59: we don't need controller anymore iirc

src/lib/tableland-calls.ts
```diff
@@ -36,44 +53,32 @@ async function parseResponse(res: any): Promise<any> {
...
 async function hash(
   this: Connection,
   query: string
-): Promise<StructureHashReceipt> {
+): Promise<StructureHashResult> {
   const message = await GeneralizedRPC.call(this, "validateCreateTable", {
     create_statement: query,
   });
-  const json = await SendCall.call(this, message);
+  if (!this.token) {
```
carsonfarmer, line 76: lazily fetch token, and lazily fetch signer, and only cache them if we update them

src/lib/token.ts
```diff
@@ -40,3 +47,19 @@ export async function createToken(
 
   return { token: token };
 }
+
+export async function userCreatesToken(
```
carsonfarmer, line 51: I'm now exposing this for use outside this sdk

src/lib/util.ts
```diff
@@ -13,43 +21,67 @@ const polyfills = {
...
-export const contractAddresses: { [index: string]: string } = {
-  // If building locally you can put your contract address here or use the contract connection option
-  localhost: "",
-  goerli: "0xa4b0729f02C6dB01ADe92d247b7425953d1DbA25",
-  "optimism-kovan": "0xf9C3530C03D335a00163382366a72cc1Ebbd39fF",
-  "polygon-mumbai": "0x70364D26743851d4FE43eCb065811402D06bf4AD",
-};
+export type NetworkName = "testnet" | "staging" | "custom";
+
+export type ChainName =
```
carsonfarmer, line 26: strict types of chains and networks now

src/lib/util.ts
```diff
@@ -13,43 +21,67 @@ const polyfills = {
...
-  {
-    key: "unknown",
+  // staging:
+  "optimism-kovan-staging": {
+    name: "optimism-kovan",
+    phrase: "Optimism Kovan",
+    chainId: 69,
+    contract: "0x322F01e81c38B4211529f334864fA630F6aeA408",
+  },
+  custom: {
```
carsonfarmer, line 68: this is just a generic placeholder for any custom network, be it localhost or otherwise

carsonfarmer:
Some potential more updates for v2 before we release. These are some big changes, but they address some issues users were having with constantly signing things. This now avoids signing until the first time a signature is required.

joewagner (approved):

PR #59, opened 2022-06-18
fix: list method returns empty array if no tables

### LINKED ISSUE #58: `list` method is returning an Object
## Expected Behavior
After connecting and calling `siwe`, calling `list` should return an Array of the callers tables.  If the caller does not have any tables, then `list` should return an empty Array.

## Actual Behavior
`list` is returning an empty Object.

## Steps to Reproduce the Problem

1. connect
2. call `siwe` and login
3. call `list` and inspect the response

This appears to be a bug in the v2 changes.  The helper `camelCaseKeys` should only be called with an Object and in the case of the `list` method is being called with an Array.

joewagner:
Fixes #58

Ran into this bug while updating the older example apps

### DIFF
--- a/src/lib/list.ts
+++ b/src/lib/list.ts
@@ -1,5 +1,5 @@
 import { TableMetadata, Connection } from "./connection.js";
-import { getSigner, camelCaseKeys } from "./util.js";
+import { getSigner } from "./util.js";
 
 export async function list(this: Connection): Promise<TableMetadata[]> {
   this.signer = this.signer ?? (await getSigner());
@@ -9,5 +9,5 @@ export async function list(this: Connection): Promise<TableMetadata[]> {
     `${this.options.host}/chain/${this.options.chainId}/tables/controller/${address}`
   ).then((r) => r.json());
 
-  return camelCaseKeys(res);
+  return res;
 }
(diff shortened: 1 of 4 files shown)


src/lib/list.ts
```diff
@@ -9,5 +9,5 @@ export async function list(this: Connection): Promise<TableMetadata[]> {
     `${this.options.host}/chain/${this.options.chainId}/tables/controller/${address}`
   ).then((r) => r.json());
 
-  return camelCaseKeys(res);
```
awmuncy, line 12: Not sure I understand _why_ this fixes the issues. Was camelCaseKeys causing the issue, was it simply not needed anymore, or both?

joewagner: Both haha,
the camelCaseKeys function uses `Object.fromEntries` and `Object.entries` to turn an Object into an iterable (Array) change the keys and then convert back to an Object.  If you pass in an Array, javascript treats the Array like an Object with keys of '0', '1', '2', etc... The function converts them to lowerCamelCase, which in that case is the same value.  This results in an Object that has keys that look like an Array, e.g. {'0': 'foo'}.
So when the tests where checking if result[0] === 'foo' they were passing.

Basically that helper should only get called with an Object. I don't know a way to imply that via Typescript aside from `any`

PR #63, opened 2022-06-30
fix #61 update tableland smart contract dependency

### LINKED ISSUE #61: Update dependency package naming
Once @tableland/eth is published with new name @tableland/evm we need to make that change here too.

joewagner:
Fixes #61

capturing the package name change of @tableland/eth to @tableland/evm


carsonfarmer (approved):
👍

PR #66, opened 2022-07-05
Get contract addresses from evm-tableland

### LINKED ISSUE #65: Deployed contract addresses listed in util.ts should be imported from evm-tableland
Capturing this after a call today.  The util.ts file tracks the addresses that the TablelandTables registry contracts are deployed to for each chain.  Soon evm-tableland will publish these addresses, and once that happens we can import them instead of manually editiing the util.ts file

joewagner:
Fixes #65

This won't be possible until evm-tableland has published a version that is exporting the addresses

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@tableland/sdk",
-  "version": "1.0.0",
+  "version": "2.0.0",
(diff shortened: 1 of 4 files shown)


carsonfarmer (approved):
Yassss!

PR #69, opened 2022-07-08
Correctly get deployed registry contract addresses

joewagner:
This fixes an issue that was preventing the sdk from connecting to the correct evm contract address


awmuncy (approved):
LGTM

Issue #67, opened 2022-07-06
Add missing evm-tableland methods


sanderpick:
Related to #56. Add support for other missing methods:

- `setController`
- `getController`
- `lockController`

PR #71, opened 2022-07-09
Initial build out of rpcRelay option

### LINKED ISSUE #70: Add enableTestnetRPCRelay flag
This flag will indicate if the smart contract interactions should go directly to the Smart Contract, or be relayed through the validator.  The validator will initially refuse to relay on any mainnet due to cost.

joewagner:
Fixes #70 

This also includes added tests for the receipt method and the resultsToObjects method

### DIFF
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -32,6 +32,10 @@ export interface StructureHashResult {
   structureHash: string;
 }
 
+export interface ValidateWriteResult {
+  tableId: number;
+}
+
 export interface CreateTableReceipt {
   tableId?: BigNumber;
   name?: string;
@@ -50,6 +54,7 @@ export interface Connection {
     chain?: ChainName;
     contract: string;
     chainId: number;
+    rpcRelay?: boolean;
   };
   list: () => Promise<TableMetadata[]>;
   create: (
--- a/src/lib/eth-calls.ts
+++ b/src/lib/eth-calls.ts
@@ -22,4 +22,23 @@ async function registerTable(
   return await tx.wait();
 }
 
-export { registerTable };
+async function runSql(
+  this: Connection,
+  tableId: number,
+  query: string
+): Promise<ContractReceipt> {
+  this.signer = this.signer ?? (await getSigner());
+  const address = await this.signer.getAddress();
+
+  const contractAddress = this.options.contract;
+
+  const contract = TablelandTables__factory.connect(
+    contractAddress,
+    this.signer
+  );
+  const tx = await contract.runSQL(address, tableId, query);
+
+  return await tx.wait();
+}
+
+export { registerTable, runSql };
--- a/src/lib/query.ts
+++ b/src/lib/query.ts
@@ -6,6 +6,7 @@
 
 import { ReadQueryResult, WriteQueryResult, Connection } from "./connection.js";
 import * as tablelandCalls from "./tableland-calls.js";
+import { runSql } from "./eth-calls.js";
 
 export function resultsToObjects({ rows, columns }: ReadQueryResult) {
   return rows.map((row: any[]) =>
@@ -24,5 +25,14 @@ export async function write(
   this: Connection,
   query: string
 ): Promise<WriteQueryResult> {
-  return await tablelandCalls.write.call(this, query);
+  if (this.options.rpcRelay) {
+    return await tablelandCalls.write.call(this, query);
+  }
+
+  // TODO: need to ask the Validator if this query is valid, and ideally it also returns the tableId
+  const { tableId } = await tablelandCalls.validateWriteQuery.call(this, query);
+
+  const txn = await runSql.call(this, tableId, query);
+
+  return { hash: txn.transactionHash };
 }
--- a/src/lib/query.ts
+++ b/src/lib/query.ts
@@ -29,7 +29,7 @@ export async function write(
     return await tablelandCalls.write.call(this, query);
   }
 
-  // TODO: need to ask the Validator if this query is valid, and ideally it also returns the tableId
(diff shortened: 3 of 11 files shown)


src/lib/connection.ts
```diff
@@ -50,6 +54,7 @@ export interface Connection {
     chain?: ChainName;
     contract: string;
     chainId: number;
+    rpcRelay?: boolean;
```
joewagner, line 57: Tried to keep the option kind of short since that is the style we've been following, but no strong opinion here

src/lib/eth-calls.ts
```diff
@@ -22,4 +22,23 @@ async function registerTable(
   return await tx.wait();
 }
 
-export { registerTable };
+async function runSql(
```
joewagner, line 25: This PR only covers the direct to smart contract call for runSQL.  If all the patterns here make sense and work well, I'm thinking we create another PR for the other smart contract methods that includes implementation for #67

src/lib/eth-calls.ts
```diff
@@ -22,4 +22,23 @@ async function registerTable(
...
+
+  const contractAddress = this.options.contract;
+
+  const contract = TablelandTables__factory.connect(
+    contractAddress,
+    this.signer
+  );
+  const tx = await contract.runSQL(address, tableId, query);
+
+  return await tx.wait();
```
joewagner, line 41: I have an Objective coming up after this to enable notifying the caller when tx receipts are materialized in the validator, but until then I just returned the receipt similar to the current `create` method.

src/lib/util.ts
```diff
@@ -48,20 +49,23 @@ export const SUPPORTED_CHAINS: Record<ChainName, SupportedChain> = {
     chainId: 5,
     contract: proxies["ethereum-goerli"],
     host: "https://testnet.tableland.network",
+    rpcRelay: true,
```
joewagner, line 52: Added a default for each tableland network.
Trying to keep it simple for now, but when we add the mainet networks I'm thinking they would all have `rpcRelay: false` until we figure out tokenomics/approved-addresses/etc...  SDK users will be able to override this, but in the case of someone trying to relay on mainnet incorrectly the validator will return an error.

test/runQuery.test.ts
```diff
@@ -1,9 +1,11 @@
 import fetch from "jest-fetch-mock";
-import { connect } from "../src/main";
+import { connect, resultsToObjects } from "../src/main";
```
joewagner, line 2: Unrelated to the rpcRelay, but I noticed we were not testing the `resultsToObjects` helper

joewagner:
NOTE: This won't work until the Validator deploys with the `tableland_validateWriteQuery` RPC endpoing https://github.com/tablelandnetwork/go-tableland/pull/169

src/lib/eth-calls.ts
carsonfarmer: Actually the create method now "unpacks" this to some degree. I think its ok to return the raw receipt here though, because we're accessing the raw smart contract method anyway.

src/lib/query.ts
```diff
@@ -24,5 +25,14 @@ export async function write(
   this: Connection,
   query: string
 ): Promise<WriteQueryResult> {
-  return await tablelandCalls.write.call(this, query);
+  if (this.options.rpcRelay) {
+    return await tablelandCalls.write.call(this, query);
+  }
+
+  // ask the Validator if this query is valid, and get the tableId for use in SC call
+  const { tableId } = await tablelandCalls.validateWriteQuery.call(this, query);
```
carsonfarmer, line 33: If I'm specifying rpcRelay false here, am I going to be surprised that I still end up making calls to the remote host in this case?

carsonfarmer: Maybe having an additional option with `validate: boolean` that defaults to `true`?

test/fauxFetch.ts
```diff
@@ -55,11 +55,35 @@ export const FetchCreateDryRunError = async () => {
   });
```
carsonfarmer, line 55: Really looking forward to being able to integrate local-tableland in our CI flow so we don't have to do these mocks.

test/runQuery.test.ts
carsonfarmer: Nice, good catch

carsonfarmer (approved):
This looks like a good first pass. I have an alternative option which I'll mention in a follow up comment. But "as is" I think this will work.

carsonfarmer:
This PR is all about making it easier to make smart contract calls "directly". One alternative option is to just make it really easy to extract the information required to instantiate an ethers.js contract object (so providing a util to get the contract address, signer, etc), that way, we don't need to individual wrap each smart contract method, and instead just expose the contract object itself. Since this is an "advanced" feature anyway, it is likely that the caller is already familiar with ethers.js contract flows.

Thoughts?

test/fauxFetch.ts
joewagner: I'm hoping to work on that this week

src/lib/eth-calls.ts
sanderpick: 👍

src/lib/query.ts
sanderpick: how about `skipQueryValidation`? imo, flags with default value `true` are confusing.

sanderpick (approved):
Awesome. I left a couple comments!

src/lib/eth-calls.ts
carsonfarmer: Yeh I think we need these in ASAP. Let's get this merged pretty soon so we can get those in subsequently.

PR #72, opened 2022-07-09
add explicit chain for local-tableland

sanderpick:
this is probably what @joewagner wanted originally... i can now see how nice it is. CLI PR coming as well.


joewagner (approved):
@sanderpick This is excellent!  It will make connection code for local development much less verbose.  When Devs start tinkering with more complex Tableland apps this kind of thing seems like it will make tutorials etc... easier to read and write

carsonfarmer (approved):
Easy one!

PR #74, opened 2022-07-12
Confirm contract events are materialized

joewagner:
### Overview
`create` and `write` methods call smart contract methods (with or without relay) which eventually materialize into table state in Validators within Tableland.  As a developer, you are currently forced to roll your own way to discover when those changes had been consumed and materialized by Validators.  The changes in this PR make the`write` and `create` methods default to waiting to resolve until the SDK's connected Validator has materialized the transaction's sql statements.

This is accomplished through simple(dirty) polling the Validator via the `receipt` method until the receipt exists or a timeout is reached.  The polling rate is calculated with exponential backoff, and then once more at the after expiration of the timeout.

### More Details
As well as defaulting to waiting for materialization, this introduces the ability to set behavior via options.
The `write` method now takes an optional second argument that is an options Object. 
The `create` method is now polymorphic.  The second argument can still be a String to be used as a table name prefix, or it can be an options Object.

Available options:
 - `skipConfirm`: Default is false, and if set to true the method returns as soon as it gets the smart contract transaction hash.  If set to true, the method waits for the Validator to confirm materialization up to `timeout`
 - `prefix`:  Only available to the `create` method.  Default is ""
 - `timeout`: Default is 120000 (2 minutes). The maximum amount of time in milliseconds the caller is willing to poll the Validator waiting to confirm the materialization of the changes.

### Notes and Thoughts
The `skipConfirm` option being set to true will probably make sense for SDK users who are on L1 mainnets.  I considered having the default value be set differently for each Tableland network, but I feel that might not make sense for devs who are working in a testnet as a way to **test** (literally) a future mainnet application.

Ref to the Objective:  https://www.notion.so/textile/Handle-receipt-confirmations-in-JS-SDK-82da86b6184f4472a42d697b5ef91f9f

### DIFF
--- a/src/lib/create.ts
+++ b/src/lib/create.ts
@@ -2,7 +2,7 @@ import { BigNumber } from "ethers";
 import { Connection, CreateTableReceipt, MethodOptions } from "./connection.js";
 import * as tablelandCalls from "./tableland-calls.js";
 import { registerTable } from "./eth-calls.js";
-import { getPrefix, shouldSkipConfirm } from "./util.js";
+import { getPrefix, getTimeout, shouldSkipConfirm } from "./util.js";
 
 /**
  * Registers an NFT with the Tableland Ethereum smart contract, then uses that to register
@@ -23,8 +23,10 @@ export async function create(
   const { chainId } = this.options;
   const prefix = getPrefix(options);
   const skipConfirm = shouldSkipConfirm(options);
+  const timeout = getTimeout(options, 120000);
 
   const query = `CREATE TABLE ${prefix}_${chainId} (${schema});`;
+
   // This "dryrun" is done to validate that the query statement is considered valid.
   // We check this before minting the token, so the caller won't succeed at minting a token
   // then fail to create the table on the Tableland network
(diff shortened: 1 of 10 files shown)


src/lib/create.ts
```diff
@@ -14,10 +15,18 @@ import { BigNumber } from "ethers";
...
   this: Connection,
   schema: string,
-  prefix: string = ""
+  // TODO: changing how this function is called would require a major version bump
+  //       making it polymophic lets us keep it in the current major verision.
+  //       when bump major remember to change this arg to only be `MethodOptions`
+  options?: MethodOptions | string
 ): Promise<CreateTableReceipt> {
   const { chainId } = this.options;
+  const prefix = getPrefix(options);
```
carsonfarmer, line 24: I like this pattern!

carsonfarmer (commented):
Looks pretty good to me here @joewagner. I think this is a pretty nice DX improvement, even if it is a bit "dirty" ;)

joewagner:
This was rebased into #79 no longer needed

PR #75, opened 2022-07-12
Update rpc methods + hash API change

### LINKED ISSUE #56: Need to add and expose a setController method
The SDK should enable easily setting the controller contract for tables.  This can be done via a direct call to the smart contract (eth-calls.ts), the validator's RPC method (tableland-calls.ts), or potentially enable both and let the SDK user choose.

### LINKED ISSUE #62: Replace "hash" function with "validateCreate" function and match "create" API
Since you don't actually write your own create statements if you are using the SDK (or the CLI), we should just match that API in the validation function as well. Additionally, hash is slightly misleading, because the name doesn't reflect what the API is really designed for... validation.

carsonfarmer:
Signed-off-by: Carson Farmer <carson.farmer@gmail.com>

Fixes #62
Fixes #56

- [x] Tests pass
- [x] Appropriate changes to README are included in PR

### DIFF
--- a/src/lib/connector.ts
+++ b/src/lib/connector.ts
@@ -5,7 +5,11 @@ import { read, write } from "./query.js";
 import { create } from "./create.js";
 import { hash } from "./hash.js";
 import { siwe } from "./siwe.js";
-import { receipt } from "./tableland-calls.js";
+import {
+  receipt,
+  setController,
+  validateWriteQuery,
+} from "./tableland-calls.js";
 import { SUPPORTED_CHAINS, NetworkName, ChainName } from "./util.js";
 import { Connection } from "./connection.js";
 
@@ -112,6 +116,12 @@ export async function connect(options: ConnectOptions): Promise<Connection> {
     },
     get siwe() {
       return siwe;
+    },
+    get setController() {
+      return setController;
+    },
+    get validate() {
+      return validateWriteQuery;
     },
   };
 
--- a/src/lib/tableland-calls.ts
+++ b/src/lib/tableland-calls.ts
@@ -16,6 +16,9 @@ export interface RpcParams {
   statement?: string;
   /* eslint-disable-next-line camelcase */
   txn_hash?: string;
+  caller?: string;
+  /* eslint-disable-next-line camelcase */
+  token_id?: string;
 }
 
 export interface RpcReceipt<T = any> {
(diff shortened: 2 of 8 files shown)


carsonfarmer:
Note that for #62 I opted _not_ to update `create` with a dryrun option, and instead just left `hash` with a matching API to `create`. That seems like a pretty useful change on its own. I'm ok also adding a dryrun option to create if we want? Additionally, note that for #56, I've only added the setController RPC method, but _not_ the other methods listed in #67 yet (getController, lockController).

src/lib/tableland-calls.ts
```diff
@@ -141,4 +144,35 @@ async function receipt(
...
 
-export { hash, list, receipt, read, validateWriteQuery, write };
+async function setController(
+  this: Connection,
+  controller: string,
+  name: string,
+  caller?: string
+): Promise<WriteQueryResult> {
+  const parts = name.split("_");
+  if (parts.length !== 3) {
```
joewagner, line 154: I don't see the arg `name` being used here, and If the prefix has an `_` the method call would fail.  
Am I missing something?

test/fauxFetch.ts
```diff
@@ -149,18 +159,19 @@ export const FetchReceiptExists = async () => {
     result: {
       receipt: {
         chainId: 5,
-        txnHash: "0xc3e7d1e81b59556f414a5f5c23760eb61b4bfaa18150d924d7d3b334941dbecd",
+        txnHash:
+          "0xc3e7d1e81b59556f414a5f5c23760eb61b4bfaa18150d924d7d3b334941dbecd",
         blockNumber: 1000,
-        tableId: '2',
```
joewagner, line 154: Nice!  Lots of cleanup needed here

src/lib/connector.ts
```diff
@@ -5,7 +5,11 @@ import { read, write } from "./query.js";
 import { create } from "./create.js";
 import { hash } from "./hash.js";
 import { siwe } from "./siwe.js";
-import { receipt } from "./tableland-calls.js";
+import {
+  receipt,
+  setController,
```
joewagner, line 10: This works well for testnets, but when we go to mainnets we will want a way to go direct to the SC.
If we create a file with a wrapper method for this, the wrapper method could determine if we want to go direct to the SC or relay through the Validator based on the SUPPORTED_CHAIN that was specified. 
Thoughts?

joewagner (changes requested):
Looks awesome! I think we should figure out the mainnet direct to SC stuff before committing.  Also, I left a comment on some confusion on the `name` arg

joewagner:
> Note that for #62 I opted _not_ to update `create` with a dryrun option, and instead just left `hash` with a matching API to `create`. That seems like a pretty useful change on its own. I'm ok also adding a dryrun option to create if we want? 

This is perfect because #74 will need the options argument adding one here would endup with a bunch of conflict.  We can add a dryrun option in another PR if that still makes sense

src/lib/tableland-calls.ts
carsonfarmer: Name is just the table name, and we want to extract the prefix from it. I guess, since we really _only_ need prefix (chainId can be derived) we could just use a prefix arg?

src/lib/connector.ts
carsonfarmer: I think that would likely be fine. In that case, would be API remain essentially the same? If yes, I'd like to get this in sooner rather than later, unless we think we can push the direct to SC stuff out soon?

src/lib/tableland-calls.ts
carsonfarmer:
```suggestion
  if (parts.length < 3) {
```

src/lib/connector.ts
joewagner: The direct to SC stuff was merged to main in #71 
I was thinking we would do something similar here

src/lib/tableland-calls.ts
joewagner: changing to `parts.length < 3` still won't end up making parts[1] be the chainId.  I was working on a fix for #76 that checks the network. Once #78 is done, maybe we could use that?

carsonfarmer: yes duh, of course, good call.

joewagner:
@carsonfarmer Is it ok to close this?  I believe everything here it made into #79

PR #78, opened 2022-07-12
Fix #76, ensure wallet and tableland share chain

### LINKED ISSUE #22: Checks on signer provider are currently too restrictive.

### LINKED ISSUE #76: Wallet can be connected to a different network than SDK
From discord: 

--------

strange, @njwags i'm getting this back from create on goerli: 
```
{
  tableId: undefined,
  prefix: 'rigs_contract',
  chainId: 5,
  txnHash: '0xd78df58512c760846f8045ded312a2ff4932b9dd912850ae40d303ab00b05653',
  blockNumber: 27146217,
  name: 'rigs_contract_5_undefined'
}
```

txn hash doesn't actually exist if you look at goerli etherscan

--------

The SDK is expecting to be connected to Goerli, but transaction was sent to [Mumbai](https://mumbai.polygonscan.com/tx/0xd78df58512c760846f8045ded312a2ff4932b9dd912850ae40d303ab00b05653).
I believe the fix will be to enforce the signer's network to match the SDK in the `siwe` method

cc: @andrewxhill @sanderpick

joewagner:
Fixes #76

This is an interesting bug. When we fixed #22 we setup the possibility that the dev can specify a tableland network, but the wallet can be connected to a different tableland network chain.

One side affect here is that the connect method is now synchronous.  This won't break our API since you can `await` anything, but basically all of our examples in the docs will be out of date.

### DIFF
--- a/src/lib/connector.ts
+++ b/src/lib/connector.ts
@@ -7,7 +7,7 @@ import { hash } from "./hash.js";
 import { siwe } from "./siwe.js";
 import { receipt } from "./tableland-calls.js";
 import { SUPPORTED_CHAINS, NetworkName, ChainName } from "./util.js";
-import { checkNetwork } from "./checkNetwork.js";
+import { checkNetwork } from "./check-network.js";
 import { Connection } from "./connection.js";
 
 /**
--- a/src/lib/connector.ts
+++ b/src/lib/connector.ts
@@ -7,6 +7,7 @@ import { hash } from "./hash.js";
 import { siwe } from "./siwe.js";
 import { receipt } from "./tableland-calls.js";
 import { SUPPORTED_CHAINS, NetworkName, ChainName } from "./util.js";
+import { checkNetwork } from "./checkNetwork.js";
 import { Connection } from "./connection.js";
 
 /**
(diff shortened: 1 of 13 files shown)


src/lib/connector.ts
```diff
@@ -48,22 +49,6 @@ export async function connect(options: ConnectOptions): Promise<Connection> {
   }
 
   const signer = options.signer;
-  if (signer && signer.provider) {
-    // Set params with provider network info if not explicitly given in options
```
joewagner, line 52: Since a signer isn't required during connection anymore, moving the check to see if the wallet and tableland chains matched to a separate method lets us do this check when tokens are created, and when direct calls to the contract are made

src/lib/siwe.ts
```diff
@@ -4,8 +4,9 @@ import { getSigner } from "./util.js";
 
 export async function siwe(this: Connection): Promise<Token> {
   this.signer = this.signer ?? (await getSigner());
-  const chainId = this.options.chainId;
 
-  this.token = await userCreatesToken(this.signer, chainId);
+  await this.checkNetwork();
```
joewagner, line 8: Waiting to check that the chains match until we create the token.

carsonfarmer: Yeh killer, more laziness!

src/lib/connector.ts
```diff
@@ -37,9 +38,9 @@ export interface ConnectOptions {
  * @param options Options to control client connection.
  * @returns Promise that resolves to a Connection object.
  */
-export async function connect(options: ConnectOptions): Promise<Connection> {
+export function connect(options: ConnectOptions): Connection {
```
carsonfarmer, line 41: This is really nice! While slightly breaking (in terms of docs/usage), I think it is going to improve the DX over the long run. Nice find/fix.

carsonfarmer (approved):
Yeh this is a really nice improvement.

PR #81, opened 2022-07-16
Joe/new methods

### LINKED ISSUE #67: Add missing evm-tableland methods

joewagner:
Fixes #67 

This adds `getController`, `setController`, and `lockController`.  The Validator RPC does not yet support these fully, but direct to contract is working.

### DIFF
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -95,6 +95,7 @@ export interface Connection {
     name: string
   ) => Promise<WriteQueryResult>;
   getController: (tableName: string) => Promise<string>;
+  lockController: (tableName: string) => Promise<WriteQueryResult>;
   siwe: () => Promise<Token>;
   validate: (query: string) => Promise<ValidateWriteResult>;
   waitConfirm: (
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -94,6 +94,7 @@ export interface Connection {
     controller: string,
     name: string
   ) => Promise<WriteQueryResult>;
+  getController: (tableName: string) => Promise<string>;
   siwe: () => Promise<Token>;
   validate: (query: string) => Promise<ValidateWriteResult>;
   waitConfirm: (
--- a/src/lib/eth-calls.ts
+++ b/src/lib/eth-calls.ts
@@ -60,4 +60,19 @@ async function setController(
   return await tx.wait();
 }
 
-export { registerTable, runSql, setController };
+async function getController(
+  this: Connection,
+  tableId: number
+): Promise<string> {
+  this.signer = this.signer ?? (await getSigner());
+
+  const contractAddress = this.options.contract;
+
+  const contract = TablelandTables__factory.connect(
+    contractAddress,
+    this.signer
+  );
+  return await contract.getController(tableId);
+}
+
+export { registerTable, runSql, setController, getController };
--- a/src/lib/eth-calls.ts
+++ b/src/lib/eth-calls.ts
@@ -75,4 +75,22 @@ async function getController(
   return await contract.getController(tableId);
 }
 
-export { registerTable, runSql, setController, getController };
+async function lockController(
+  this: Connection,
+  tableId: number
+): Promise<ContractReceipt> {
+  this.signer = this.signer ?? (await getSigner());
+  const caller = await this.signer.getAddress();
+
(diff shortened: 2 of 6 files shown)


src/lib/connection.ts
```diff
@@ -94,6 +94,8 @@ export interface Connection {
     controller: string,
     name: string
   ) => Promise<WriteQueryResult>;
+  getController: (tableName: string) => Promise<string>;
```
carsonfarmer, line 97: Just wondering out loud... should we namespace all of these under controller? so controller.get, controller.set, controller.lock?

src/lib/eth-calls.ts
```diff
@@ -60,4 +60,37 @@ async function setController(
...
+
+  const contractAddress = this.options.contract;
+
+  const contract = TablelandTables__factory.connect(
+    contractAddress,
+    this.signer
+  );
+  const tx = await contract.lockController(caller, tableId);
+
+  return await tx.wait();
```
carsonfarmer, line 93: Why diff from above. Just because this is an actual mutating call?

src/lib/get-controller.ts
```diff
@@ -0,0 +1,27 @@
+import { Connection } from "./connection.js";
+//import * as tablelandCalls from "./tableland-calls.js";
```
carsonfarmer, line 2: Causing a lint error. Maybe just remove for now?

src/lib/get-controller.ts
```diff
@@ -0,0 +1,27 @@
...
+export async function getController(
+  this: Connection,
+  tableName: string
+): Promise<string> {
+  const tableId = tableName.trim().split("_").pop();
+  if (typeof tableId !== "string") throw new Error("malformed tablename");
+
+  // TODO: do we need to enable rpcRelay?
+  //       It's a read so I don't see any point aside from having a more robust rpc api
+  //if (this.options.rpcRelay) {
```
carsonfarmer, line 18: Linter doesn't like this.

src/lib/tableland-calls.ts
```diff
@@ -38,6 +38,10 @@ export interface SetControllerParams {
   token_id: string;
 }
 
+export interface LockControllerParams {
+    token_id: string;
```
carsonfarmer, line 42: Needs a linter escape

carsonfarmer (commented):
Looking good so far @joewagner. Obvi some linter issues, and we'll need to decide if the RPC api needs to support these or not?

src/lib/eth-calls.ts
joewagner: Yeah, since getController is a `view` it returns a string. setController and lockController return transactions, and we've been returning the hash in those cases.

src/lib/connection.ts
joewagner: That would work.  My tendency is usually to try to keep the api intuitive to a developer when possible. Currently the Contract functions, RPC endpoints, and sdk methods all match.  This makes sense to me, but maybe a `controller` object with functions on it would be better?
If we do switch, i think we would want to make sure to bind the context of the functions to "tableland", e.g. inside of tableland.controller.set we would want `this` to be tableland not controller so we can call `this.checkNetwork` etc...  Alternatively we could abstract the idea of a `super` as `this.super` (or something like that) that points to tableland, but that seems more complicated.

joewagner:
@carsonfarmer sorry about all the linting issues.  Was kinda stuck until we decided if we want to have the RCP server relay all these methods.  
We have been discussing having the RPC server relay creating tables.  If we are doing that, it seems like these methods probably make sense too.  This winter when we discussed relay vs. direct it seemed like we wanted to mirror what the network might do once we work out all the validator incentives etc...  Should we do a TPD for what gets relayed and what doesn't?

carsonfarmer (approved):
Sorry for the delay here @joewagner things have been hectic. Don't wanna hold this one up any longer though! Looks good to me.

joewagner:
> Sorry for the delay here @joewagner things have been hectic. Don't wanna hold this one up any longer though! Looks good to me.

No problem!  Definitely a lot of more important things happening recently...
I'll merge this now, do we have plans for when the next release will be?  I think some of the pilot program folks could use some of the features that we've built out recently.

PR #88, opened 2022-08-02
Add methods for schema/ and structure/ routes

joewagner:
Add methods for the discovery RPC endpoints.
See https://github.com/tablelandnetwork/go-tableland/pull/208 for more detail

### DIFF
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -9,11 +9,13 @@ export interface TableMetadata {
   structure: string;
 }
 
-export type Columns = Array<{
+export type Columns = Array<{ name: string }>;
+export type SchemaColumns = Array<{
   name: string;
   type?: string;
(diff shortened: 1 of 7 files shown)


src/lib/connection.ts
```diff
@@ -9,7 +9,11 @@ export interface TableMetadata {
   structure: string;
 }
 
-export type Columns = Array<{ name: string }>;
+export type Columns = Array<{
```
carsonfarmer, line 12: Do we want to have a separate columns type for query responses vs the responses _with_ type and constraint data?

carsonfarmer (approved):
Lovely, short and sweet!

src/lib/connection.ts
joewagner: Good call, I'll update now

PR #89, opened 2022-08-05
Release v3.0.0

joewagner:
Merge release into main.
Now that we don't allow direct pushes to main I wasn't able to do `npm version ...` then `git push origin main`.  When we implement changes needed for #68 this kind of pr shouldn't be needed anymore

PR #90, opened 2022-08-09
Better typing, more consistent relay options

joewagner:
This is a fix to an issue from Discord: https://discord.com/channels/592843512312102924/1006245128589037619
See below if you don't have Discord access.

TL;DR; The rpcRelay option is potentially specified in the sdk connect method, but devs have been wanting to specify with a per method call option.  We had started to implement this, but never finished.  
This also makes the types more specific.

**OP from Discord:**
> issue re: rpcRelay and potential missing impl...or we just need better docs on MethodOptions since its rpcRelay is an optional param that is not used at all in certain methods. OR -- we havent implemented the method-level logic to process MethodOptions and its rpcRelay yet
> https://discord.com/channels/592843512312102924/968582612945870878/1006196867765391391
> 
> after digging into this a bit, it looks like write can take a MethodOptions param as part of the Connection object's definition. but, if you look at the write method's impl itself, there is not a param to pass the MethodOptions. thus, when a user tries to pass rpcRelay: false in a write method call, the default Connection's rpcRelay is used (which is likely true unless otherwise specified). the Connection's rpcRelay logic is also linked below, for reference, as there might be some other aspects to consider, as noted in the original Discord comment.
> 
> - Connection showing the options can be passed to write: https://github.com/tablelandnetwork/js-tableland/blob/ba13ae3355971403801d758a03963e94f2a6c189/src/lib/connection.ts#L107
> - write, which doesnt take MethodOptions at all: https://github.com/tablelandnetwork/js-tableland/blob/ba13ae3355971403801d758a03963e94f2a6c189/src/lib/tableland-calls.ts#L143
> - create, which doesn't use rpcRelay from the options. i suppose this makes sense from a logic perspective b/c we dont give the option to relay creates at the validator...but we do offer validator relays for writes, which i think is a conflicting DX across these two calls. note that hash has a similar situation: https://github.com/tablelandnetwork/js-tableland/blob/ba13ae3355971403801d758a03963e94f2a6c189/src/lib/create.ts#L19
> - rpcRelay logic within Connection: https://github.com/tablelandnetwork/js-tableland/blob/ba13ae3355971403801d758a03963e94f2a6c189/src/lib/connector.ts#L74
> 
> also, write only has logic to use relayWriteQuery under the hood and doesnt use runSql (from eth-calls.ts), so it looks like we'd have to add some logic to make a direct SC call for write as well? 
> 
> there's a bunch going on that i dont have the best insight into from an engg perspective, so i just wanna make sure my logic is correct before i open a GH issue or formalize this as a friction log. maybe its all in an obj somewhere that i couldnt find
>

### DIFF
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -70,6 +70,25 @@ export interface MethodOptions {
   timeout?: number;
 }
 
+export interface CreateOptions {
+  prefix?: string;
+  skipConfirm?: boolean;
+  timeout?: number;
+}
+
+export interface WriteOptions {
+  skipConfirm?: boolean;
+  rpcRelay?: boolean;
+}
+
+export interface SetControllerOptions {
+  rpcRelay?: boolean;
+}
+
+export interface HashOptions {
+  prefix?: string;
+}
+
 export interface ConfirmOptions {
   timeout?: number;
   rate?: number;
@@ -84,7 +103,7 @@ export interface Connection {
     chain?: ChainName;
     contract: string;
     chainId: number;
-    rpcRelay?: boolean;
+    rpcRelay: boolean;
   };
   list: () => Promise<TableMetadata[]>;
   create: (
@@ -99,21 +118,16 @@ export interface Connection {
     /**
      *  an optional options argument to specify conditions of create.
      **/
-    options?: MethodOptions
+    options?: CreateOptions
   ) => Promise<CreateTableReceipt>;
   read: (query: string) => Promise<ReadQueryResult>;
-  write: (
-    query: string,
-    options?: MethodOptions | undefined
-  ) => Promise<WriteQueryResult>;
-  hash: (
-    schema: string,
-    options?: MethodOptions
-  ) => Promise<StructureHashResult>;
+  write: (query: string, options?: WriteOptions) => Promise<WriteQueryResult>;
+  hash: (schema: string, options?: HashOptions) => Promise<StructureHashResult>;
   receipt: (txnHash: string) => Promise<ReceiptResult | undefined>;
   setController: (
     controller: string,
-    name: string
+    name: string,
+    options?: SetControllerOptions
   ) => Promise<WriteQueryResult>;
   getController: (tableName: string) => Promise<string>;
   lockController: (tableName: string) => Promise<WriteQueryResult>;
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -63,13 +63,6 @@ export interface StructureQueryResult {
   structure: string;
 }
 
-export interface MethodOptions {
-  prefix?: string;
-  skipConfirm?: boolean;
-  rpcRelay?: boolean;
-  timeout?: number;
-}
-
 export interface CreateOptions {
   prefix?: string;
   skipConfirm?: boolean;
@@ -93,6 +86,11 @@ export interface ConfirmOptions {
   timeout?: number;
   rate?: number;
 }
+
+export type PrefixOptions = HashOptions | CreateOptions;
+export type RelayOptions = WriteOptions | SetControllerOptions;
+export type SkipConfirmOptions = WriteOptions | CreateOptions;
+export type TimeoutOptions = ConfirmOptions | CreateOptions;
 
 export interface Connection {
   token?: Token;
--- a/src/lib/set-controller.ts
+++ b/src/lib/set-controller.ts
@@ -1,7 +1,12 @@
-import { Connection, WriteQueryResult } from "./connection.js";
+import {
+  Connection,
+  WriteQueryResult,
+  SetControllerOptions,
+} from "./connection.js";
 import * as tablelandCalls from "./tableland-calls.js";
 import * as ethCalls from "./eth-calls.js";
 import { checkNetwork } from "./check-network.js";
+import { shouldRelay } from "./util.js";
 
 /**
  * Set the Controller contract on a table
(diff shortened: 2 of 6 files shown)


src/lib/connection.ts
```diff
@@ -70,6 +70,25 @@ export interface MethodOptions {
   timeout?: number;
 }
 
+export interface CreateOptions {
```
joewagner, line 73: the `MethodOptions` interface was used for all of these, which was confusing at best.  Splitting these into their own interfaces seems to make more sense and be better/easier for typescript devs

src/lib/set-controller.ts
```diff
@@ -10,12 +15,14 @@ import { checkNetwork } from "./check-network.js";
...
-  name: string
+  name: string,
+  options?: SetControllerOptions
 ): Promise<WriteQueryResult> {
   const tableId = name.trim().split("_").pop();
   if (typeof tableId !== "string") throw new Error("malformed tablename");
 
-  if (this.options.rpcRelay) {
+  const doRelay = shouldRelay(this, options);
+  if (doRelay) {
```
joewagner, line 25: This allows the caller to specify if they want to use the relay on a per call basis.

src/lib/util.ts
```diff
@@ -169,6 +169,18 @@ export function shouldSkipConfirm(options?: MethodOptions): boolean {
   return !!options.skipConfirm;
 }
 
+export function shouldRelay(
```
joewagner, line 172: Use the same logic to decide if caller wants to relay for all methods

joewagner:
**Note:** we should wait until after #89 is merged then point this at `main`.

I've been thinking about the idea of relay vs. direct in 2 different ways.
1. If the Validator supports the relay, always do it if possible.  This boils down to mainnet vs. testnet.
2. If you are a dev building an app. You might start on testnet then switch to mainnet when finished.  If you want to try for consistent behavior between testnet and mainnet you might always want to go direct since the Validator isn't going to pay for things on mainnet.

This change gives more flexibility to the dev, but keeps our current default, i.e. number 1 above.

src/lib/util.ts
dtbuchholz: @joewagner Random TS question as I'm still learning the syntax/intricacies.

Since `shouldRelay` can take `WriteOptions` or `SetControllerOptions` as well, does it make sense to use a "discriminated union"? Idk if that makes things more complicated and/or isn't really necessary in TS world.

E.g., something like an `Options` type is the param in `shouldRelay`, instead of  `MethodOptions`. Or maybe this is what `MethodOptions` could be? Along these lines:
```
type Options = WriteOptions | SetControllerOptions | MethodOptions
```

dtbuchholz (approved):
Lgtm! Left a comment re: `shouldRelay` param types. Not sure if it's needed, though, so approving.

Disclaimer: this is the first PR I've reviewed here, in case another person wants to take a look / approve, too. Otherwise, nbd. Also, lmk if this wasn't a best practice.

sanderpick (approved):
Nice, LGTM!

src/lib/util.ts
joewagner: @dtbuchholz Good call! I can update this unless you want to get in on this PR?  If so, feel free to commit to this branch.  Otherwise I'll update this tonight

dtbuchholz: Sure, I can take a stab

dtbuchholz: So, I ended up adding a few different discriminated unions. I'm not sure if this adds unnecessary types for the sake of readability.

For example, if the union is `type RelayOptions = WriteOptions | SetControllerOptions`, instead, this could be done in the function signature -- e.g., `options: RelayOptions` could instead be `options: WriteOptions | SetControllerOption`. No strong preference here, not sure what's a better practice.

dtbuchholz: FYI, `npm i` from both `main` and this PR's branch throws some error, complaining about mismatch with `ethers` and `siwe`. It works if I use the `--force` flag, though, so I figured I'd share for visibility purposes:
```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
npm ERR!
npm ERR! While resolving: siwe@1.1.6
npm ERR! Found: ethers@5.6.9
npm ERR! node_modules/ethers
npm ERR!   ethers@"^5.5.2" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer ethers@"5.5.1" from siwe@1.1.6
npm ERR! node_modules/siwe
npm ERR!   siwe@"^1.1.6" from the root project
npm ERR!
npm ERR! Conflicting peer dependency: ethers@5.5.1
npm ERR! node_modules/ethers
npm ERR!   peer ethers@"5.5.1" from siwe@1.1.6
npm ERR!   node_modules/siwe
npm ERR!     siwe@"^1.1.6" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.
npm ERR!
npm ERR! See /Users/db/.npm/eresolve-report.txt for a full report.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/db/.npm/_logs/2022-08-12T00_07_04_250Z-debug-0.log
```

src/lib/connection.ts
```diff
@@ -94,6 +87,11 @@ export interface ConfirmOptions {
   rate?: number;
 }
 
+export type PrefixOptions = HashOptions | CreateOptions;
```
dtbuchholz, line 90: Removed `MethodOptions` and opted for various discriminated unions that align to a corresponding `util.ts` function.

carsonfarmer:
Hey @dtbuchholz looks like your commit here wasn't signed. Do you have that setup on your end? If not, let's sync on this. In the mean time, @joewagner looks like you can't merge until we get that sorted (we have this repo set to not allow admin overrides).

joewagner:
>  @joewagner looks like you can't merge until we get that sorted (we have this repo set to not allow admin overrides).

I think that's right, but no hurry here.

carsonfarmer:
Excellent. Thanks @dtbuchholz! @joewagner we're good to go here, merge when ready.

PR #99, opened 2022-08-16
add hack for polygon underpriced txns

### LINKED ISSUE #96: Alchemy Polygon provider returns underpriced gasPrice
Some setup to show the issue inside `registerTable`:

```
async function registerTable(
  this: Connection,
  query: string
): Promise<ContractReceipt> {
  this.signer = this.signer ?? (await getSigner());
  const address = await this.signer.getAddress();

  const contractAddress = this.options.contract;

  const contract = TablelandTables__factory.connect(
    contractAddress,
    this.signer
  );

  const oldGasPrice = await this.signer.getGasPrice();
  console.log("oldGasPrice", oldGasPrice.toNumber());

  const feeData = await this.signer.getFeeData();
  console.log("gasPrice", feeData.gasPrice?.toNumber());
  console.log("maxFeePerGas", feeData.maxFeePerGas?.toNumber());
  console.log("maxPriorityFeePerGas", feeData.maxPriorityFeePerGas?.toNumber());

  const tx = await contract.createTable(address, query);

  return await tx.wait();
}
```

This issue (https://github.com/ethers-io/ethers.js/discussions/2439#discussioncomment-1857403) is unrelated, but has a relevant comment that `provider.getGasPrice` should no longer be used. However, on both Optimism and Polygon, both `provider.getGasPrice` and `feeData.gasPrice` are the same. Following that commenter's logic, we get a giant (too big `maxFeePerGas`).

In any case, at the bottom of the thread is a comment from somebody else saying that they must always specify `gasPrice`. I suspect they are hitting the same issue we are hitting.

The `gasPrice` returned by Alchemy (using either method mentioned above) is too low compared to what metamask uses (see https://gas-api.metaswap.codefi.network/networks/137/suggestedGasFees) or what is reported on Polyscan: https://polygonscan.com/gastracker.

If I bump the Alchemy-suggested `gasPrice` by 10%, I can create tables:

```
async function registerTable(
  this: Connection,
  query: string
): Promise<ContractReceipt> {
  this.signer = this.signer ?? (await getSigner());
  const address = await this.signer.getAddress();

  const contractAddress = this.options.contract;

  const contract = TablelandTables__factory.connect(
    contractAddress,
    this.signer
  );

  const oldGasPrice = await this.signer.getGasPrice();
  console.log("oldGasPrice", oldGasPrice.toNumber());

  const feeData = await this.signer.getFeeData();
  console.log("gasPrice", feeData.gasPrice?.toNumber());
  console.log("maxFeePerGas", feeData.maxFeePerGas?.toNumber());
  console.log("maxPriorityFeePerGas", feeData.maxPriorityFeePerGas?.toNumber());

  let gasPrice;
  if (feeData.gasPrice) {
    gasPrice = Math.floor(feeData.gasPrice.toNumber() * 1.1);
  }
  console.log("bumped gasPrice", gasPrice);

  const tx = await contract.createTable(address, query, {
    gasPrice,
  });

  return await tx.wait();
}
```

Perhaps we should query the metamask or polyscan API for suggested `gasPrice` instead of relying on the provider (Alchemy in my case).

sanderpick:
Fixes #96

> It's a good idea to open an issue first for discussion.

- [x] Tests pass
- [x] Appropriate changes to README are included in PR

### DIFF
--- a/src/lib/eth-calls.ts
+++ b/src/lib/eth-calls.ts
@@ -1,4 +1,4 @@
-import { ContractReceipt } from "ethers";
+import { Overrides, ContractReceipt, Signer } from "ethers";
 /* eslint-disable-next-line camelcase */
 import { TablelandTables__factory } from "@tableland/evm";
 import { Connection } from "./connection.js";
@@ -17,7 +17,9 @@ async function registerTable(
     contractAddress,
     this.signer
   );
-  const tx = await contract.createTable(address, query);
+
+  const opts = await getOverrides(this.signer);
+  const tx = await contract.createTable(address, query, opts);
 
   return await tx.wait();
 }
@@ -36,7 +38,9 @@ async function runSql(
     contractAddress,
     this.signer
   );
-  const tx = await contract.runSQL(address, tableId, query);
+
+  const opts = await getOverrides(this.signer);
+  const tx = await contract.runSQL(address, tableId, query, opts);
 
   return await tx.wait();
 }
@@ -55,7 +59,9 @@ async function setController(
     contractAddress,
     this.signer
   );
-  const tx = await contract.setController(caller, tableId, controller);
+
+  const opts = await getOverrides(this.signer);
+  const tx = await contract.setController(caller, tableId, controller, opts);
 
   return await tx.wait();
 }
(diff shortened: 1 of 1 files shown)


src/lib/eth-calls.ts
```diff
@@ -88,9 +94,25 @@ async function lockController(
...
   );
-  const tx = await contract.lockController(caller, tableId);
+
+  const opts = await getOverrides(this.signer);
+  const tx = await contract.lockController(caller, tableId, opts);
 
   return await tx.wait();
 }
 
+async function getOverrides(signer: Signer): Promise<Overrides> {
```
carsonfarmer, line 104: Nice, I quite like how you did this, pretty non-invasive, and easy to rip out if price estimates get back on track downstream. Just to be certain, it looks like if signer _doesn't_ have a provider this will just skip along and return empty opts (which is good), correct?

carsonfarmer (approved):
LGTM. Just one clarifying question (I love these small, targeted PRs), but I'm pretty sure the answer is "yup" and we're GTG.

src/lib/eth-calls.ts
sanderpick: yep, but btw, looks like https://github.com/tablelandnetwork/js-tableland/blob/main/src/lib/check-network.ts#L15 enforces provider is always set by the time we get to the contract calls.

carsonfarmer: Ah yeh, I wasn't paying attention. This is in `eth-calls`... so we do actually need a provider here always anyway 👍 .

carsonfarmer:
Merge away!

PR #105, opened 2022-08-16
rename registerTable to createTable

### LINKED ISSUE #97: Rename registerTable to createTable
We don't use `registerTable` anywhere else in Tableland, may as well change it here

joewagner:
fix #97 Rename the registerTable to createTable.
Since registerTable was internal this is a pretty simple change


carsonfarmer (approved):
Easy peasy 👍

PR #111, opened 2022-08-17
Update connect to accept an optional siweUri argument

datadanne:
URI is required in a `SiweMessage` according to the [siwe spec](https://eips.ethereum.org/EIPS/eip-4361) so we need a way to pass in a URI when creating messages in non-browser environments, like the cli. This is also required to upgrade siwe to v2 since `SiweMessage` validates that the uri field is a valid URI here https://github.com/spruceid/siwe/blob/07103f2b72cfe0363d5c605cd6ebf9791d82cad5/packages/siwe/lib/client.ts#L382-L388

Do you think using a connection option is the right way to do it? If yes, do you prefer to use a default fallback value (like I did with tableland.xyz) or do you think `connect` should throw an error if a siwe URI isn't provided? I was debating between the two options but went with not requiring it since you can use the read only function sin the sdk without needing to siwe

### DIFF
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -102,6 +102,7 @@ export interface Connection {
     contract: string;
     chainId: number;
     rpcRelay: boolean;
+    siweUri: string;
   };
   list: () => Promise<TableMetadata[]>;
   create: (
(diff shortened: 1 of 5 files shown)


carsonfarmer:
Yes I think your intuition is right here. The other option is to provide no default, and lazily fail if/when a uri is needed but not provided. However, I think defaulting to tableland.xyz works because there is only one active gateway at the moment anyway.

datadanne:
Closing this in favor of https://github.com/tablelandnetwork/js-tableland/pull/112

Issue #106, opened 2022-06-21
ethers version conflict 


sanderpick:
<img width="645" alt="Screen Shot 2022-06-21 at 11 41 00 AM" src="https://user-images.githubusercontent.com/361000/174874388-e389ed84-01d8-4fbc-bbff-199376ad8368.png">

carsonfarmer:
Looks like this update has been merged and released. Let's update the siwe version we're using here.

PR #112, opened 2022-08-17
Update siwe to v2

datadanne:
Updates siwe to v2. I had to update the tests to use a valid address since the address is now validated when you create a  `SiweMessage`, and siwe started using `ethers.utils.getAddress` so I had to add that to the ethers mock module.

This PR contains the changes from #111 as well since that or another solution to providing a uri to `SiweMessage` is needed. 
   
Closes https://github.com/tablelandnetwork/js-tableland/issues/106
~~Blocked by https://github.com/tablelandnetwork/js-tableland/pull/111~~


datadanne:
> This is great. I'm pretty happy with the minimal updates required to get this going. You could even just merge #111 into this PR if you'd like, because it is unlikely one is going in without the other.

Sweet! Ok, this PR contains all commits from #111 so I can just close that one and mark this as ready.

.github/workflows/docs.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Setup ⬢
         uses: actions/setup-node@v2-beta
         with:
-          node-version: 14
+          node-version: 16
```
carsonfarmer, line 30: Nice to be able to do this now!

package.json
```diff
@@ -71,13 +71,15 @@
     "add-js-extension": "^1.0.4",
     "babel-core": "^7.0.0-bridge.0",
     "babel-jest": "^27.4.6",
+    "buffer": "^6.0.3",
```
carsonfarmer, line 74: Is this _required_ I'd really rather avoid it if possible?

package.json
```diff
@@ -71,13 +71,15 @@
...
     "esbuild": "^0.15.3",
     "eslint": "^8.19.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-config-standard": "^17.0.0",
     "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-n": "^15.0.0",
     "eslint-plugin-promise": "^6.0.0",
+    "events": "^3.3.0",
```
carsonfarmer, line 82: same with this one...

src/lib/connector.ts
```diff
@@ -40,6 +40,8 @@ export interface ConnectOptions {
   contract?: string;
   // Boolean that indicates if tableland writes should be relayed via Validator
   rpcRelay?: boolean;
+  // SIWE URI. Defaults to document.location.origin when used in a browser environment and to https://tableland.xyz if document.location is not available.
+  siweUri?: string;
```
carsonfarmer, line 44: 👍

src/lib/connector.ts
```diff
@@ -75,6 +77,10 @@ export function connect(options: ConnectOptions): Connection {
     typeof options.rpcRelay === "boolean" ? options.rpcRelay : info.rpcRelay;
   // If a token was provided, we cache it
   const token = options.token;
+  const siweUri =
+    options.siweUri ??
+    globalThis.document?.location.origin ??
+    "https://tableland.xyz";
```
carsonfarmer, line 83: Wondering if this should be testnet.tableland.xyz? cc @joewagner here as well. Now that I'm seeing this again, perhaps we actually _do_ require it be specified explicitly?

test/connect.test.ts
```diff
@@ -94,6 +96,24 @@ describe("connect function", function () {
     expect(connection1.token?.token === connection2.token?.token).toBe(true);
   });
 
+  test("allows specifying SIWE URI", async function () {
```
carsonfarmer, line 99: Yes pls thanks for the test.

carsonfarmer (approved):
I'm good to see this merged. But I would like to see either @sanderpick and/or @joewagner take a quick gander as well.

src/lib/connector.ts
sanderpick: testnet.tableland.network sounds right to me

sanderpick (approved):
lgtm! just one nit about the default origin

joewagner:
This has some widespread implications in the Validator as well as the Go client, so before we commit to anything it would be good to get an opinion from @jsign and/or @brunocalza 

TL;DR; I think we might have got this wrong (my fault) when we built the siwe functionality originally, and we should probably triple check we are following the spec correctly.

Details:
From the spec https://eips.ethereum.org/EIPS/eip-4361: 

> Sign-In with Ethereum works as follows:
> 
> 1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the [EIP-191](https://eips.ethereum.org/EIPS/eip-191) signed data format. Before signing, the message is prefixed with \x19Ethereum Signed Message:\n<length of message> as defined in [EIP-191](https://eips.ethereum.org/EIPS/eip-191). The message MUST incorporate an Ethereum address, domain requesting the signing, version of the message, a chain identifier chain-id, uri for scoping, nonce acceptable to the relying party, and issued-at timestamp.
> ...

Specifically there are two fields, `domain` and `uri`.  Both are required.
 - domain is: "the RFC 3986 authority that is requesting the signing".  Reading the spec again, I'm interpreting this as the website https://example.xyz is requesting that you sign this message.  I think we want to set this based on `document.location.origin`, or let it be specified during connection for the cases of node.js env.  This is currently "Tableland", which is not a valid RFC 3986 so definitely need to change this.
 - uri is: "an RFC 3986 URI referring to the resource that is the subject of the signing (as in the subject of a claim)."  This would be the value of `host`, which defaults based on the Tableland Network in `utils.ts` or can be set during connection if there is a custom Validator/Gateway being used.

If we implement changes here they will also have to be implemented in the validator, i.e. https://github.com/tablelandnetwork/go-tableland/blob/main/internal/router/middlewares/authentication.go#L77 and https://github.com/tablelandnetwork/go-tableland/blob/main/pkg/siwe/siwe.go#L25-L27
For the authentication maybe we could get the URI of the Gateway being served in the context of the request as the value of `uri`.  Then instead of checking `domain` we check that the siwe token`uri` value matches the Gateway URI.  
@jsign @brunocalza double checking, I am thinking about this right?

cc: @sanderpick @carsonfarmer Does this all make sense to you as well?

package.json
datadanne: The reason why I added them is that `siwe-parser` depends on `keccak` which depends on `readable-stream` which requires events and buffer but doesn't specify them as dependencies, and esbuild doesn't automatically include browser compatible versions of node modules when bundling.

But it depends on what the goal with the esbuild output is, what is it currently used for? I don't think I know enough about esbuild or npm modules to figure that out from the code, I don't see `dist/esbuild` referenced anywhere in `package.json`. If the bundle is only meant to be used in node environments we could pass `--platform=node` to `esbuild` instead of adding them as dev dependencies. 

Releasing code that works in both node and browser environments is kind of annoying.. I did some more reading and I think that unless we include buffer and events as 'real' dependencies in package.json (I added them as devDependencies) the sdk won't work in browsers unless a bundler is being used that includes those modules by default (like webpack and browserify). And on top of that, I realized that `keccak` relies on a global Buffer (keccak doesn't require/import it) so a browser build will have to provide that as well unless we update the sdk with something like
```ts
import { Buffer } from "buffer";

if (!globalThis.Buffer) globalThis.Buffer = Buffer;
```

jsign:
> This has some widespread implications in the Validator as well as the Go client, so before we commit to anything it would be good to get an opinion from @jsign and/or @brunocalza
> 
> TL;DR; I think we might have got this wrong (my fault) when we built the siwe functionality originally, and we should probably triple check we are following the spec correctly.
> 
> Details: From the spec https://eips.ethereum.org/EIPS/eip-4361:
> 
> > Sign-In with Ethereum works as follows:
> > 
> > 1. The wallet presents the user with a structured plaintext message or equivalent interface for signing with the [EIP-191](https://eips.ethereum.org/EIPS/eip-191) signed data format. Before signing, the message is prefixed with \x19Ethereum Signed Message:\n as defined in [EIP-191](https://eips.ethereum.org/EIPS/eip-191). The message MUST incorporate an Ethereum address, domain requesting the signing, version of the message, a chain identifier chain-id, uri for scoping, nonce acceptable to the relying party, and issued-at timestamp.
> >    ...
> 
> Specifically there are two fields that I think we should update, `domain` and `uri`. Both are required.
> 
> * domain is: "the RFC 3986 authority that is requesting the signing".  Reading the spec again, I'm interpreting this as the website https://example.xyz is requesting that you sign this message.  I think we want to set this based on `document.location.origin`, or let it be specified during connection for the cases of node.js env.  This is currently "Tableland", which is not a valid RFC 3986 so definitely need to change this.
> * uri is: "an RFC 3986 URI referring to the resource that is the subject of the signing (as in the subject of a claim)."  This would be the value of `host`, which defaults based on the Tableland Network in `utils.ts` or can be set during connection if there is a custom Validator/Gateway being used.
> 
> If we implement changes here they will also have to be implemented in the validator, i.e. https://github.com/tablelandnetwork/go-tableland/blob/main/internal/router/middlewares/authentication.go#L77 and https://github.com/tablelandnetwork/go-tableland/blob/main/pkg/siwe/siwe.go#L25-L27 For the authentication maybe we could get the URI of the Gateway being served in the context of the request as the value of `uri`. Then instead of checking `domain` we check that the siwe token`uri` value matches the Gateway URI. @jsign @brunocalza double checking, I am thinking about this right?
> 
> cc: @sanderpick @carsonfarmer Does this all make sense to you as well?

Makes sense to me. Any of these changes in the validator are simple. We already have a config attribute in the validator with the DNS where the gateway is served; so we can validate with that.

What I think we might need to think a bit more about is how all this makes sense when there're more validators and gateways.
- These values would have defaults that users can change with SDK options?
- What does it mean to validate a DNS in `uri` if a validator doesn't have a DNS?
- We should make optional the `uri` checking in the validator to solve the above bullet?

Finally, doing this kind of change in a single version will force people to immediately update to the latest SDK version which might be a bit violent. We could do the change in the validator to support "both formats/validations" for a while to at least give people a week or similar before their requests start being rejected.

sanderpick:
> * These values would have defaults that users can change with SDK options?

That's a good question. It feels natural to allow an SDK user (and downstream CLI user) to specify their validator of choice. I suspect we'll hit this need as soon as we have multiple validators (soon!) cc @joewagner, could you create an issue for this where we can discuss it? 

> * What does it mean to validate a DNS in `uri` if a validator doesn't have a DNS?

Good question. We could use IP address but that might look scary to the user. We could use a "nickname" that the user gives their validator of choice in the case where it doesn't have a DNS. At least this way it would look familiar and be easy to cross check visually. 

> * We should make optional the `uri` checking in the validator to solve the above bullet?

Sounds right

> Finally, doing this kind of change in a single version will force people to immediately update to the latest SDK version, which might be a bit violent. We could do the change in the validator to support "both formats/validations" for a while to at least give people a week or similar before their requests start being rejected. After a ~week, we can remove the old rules and only leave the new ones.

Nice, sounds right!

joewagner:
> It feels natural to allow an SDK user (and downstream CLI user) to specify their validator of choice. I suspect we'll hit this need as soon as we have multiple validators (soon!) cc @joewagner, could you create an issue for this where we can discuss it?

@sanderpick I created #117 and included some of the outstanding questions from the above discussion.

sanderpick:
Perfect, thank you

sanderpick:
How we doing here guys? Any open questions that need addressing?

sanderpick:
I'm actually not sure what changes are needed elsewhere. @joewagner / @carsonfarmer probably has the best bird's eye view of it. can you guys create some issues on other repos that this PR can be blocked by?

carsonfarmer:
I'm going to defer to @joewagner on this one as he is most aware of the work required here. @joewagner I can support you if you want to tag me on stuff.

joewagner:
From a birds eye view, I think we can separate this into two issues.
1. Our SIWE message is not being constructed and validated in a way that matches the spec.  I don't know if anything is really broken because of this, but it feels to me like something we should fix soon for the sake of following best practices.
2. Our npm siwe dependency version needs to be updated to enable running `npm install` in this repo without the force flag.  

@datadanne has an excellent PR to fix number 2. I ran all the integration tests and this can be merged without breaking anything or making any changes in the validator. 

@sanderpick @jsign @carsonfarmer If everyone is on board, I'll suggest we merge this then I can create issues that describe what we need to do to fix number 1 here and in go-validator.  Give a 👍 if this makes sense.
Alternatively, we could make a plan to follow the siwe spec more closely and create blocking issues for this PR. Give a 👎  if this makes sense.

joewagner (approved):
After the discussion it seems like merging this and addressing the message format changes in another pr makes the most sense.

carsonfarmer:
Agreed, let's go for it.

joewagner:
> TL;DR: Users of the SDK will have to make sure that an implementation of Buffer is available in the global scope, which it isn't in browsers by default

@datadanne Thanks for the reminder. Would you be able to create an issue for this and assign me?  Sounds like an easy fix, but I don't want to forget about it since this is already merged

joewagner:
Ok, good to know. Thanks!

Issue #121, opened 2022-08-24
calling the `write` method with valid sql that violates a table constraint should throw an error when the caller is waiting for confirmation


joewagner:
**Describe the bug**
When you create a table with a primary key, then try to insert duplicate values for that primary key using the `write` method, the method does not throw an error even though the writes do not succeed.

**To Reproduce**
1. Create a table with a primary key.
2. Insert the same value for the primary key twice.

**Expected behavior**
Assuming the caller is using the default value for `skipConfirm` an error should be thrown after the second insert fails.

See chain of messages in Discord for more detail: https://discord.com/channels/592843512312102924/902337528902860820/1012107267816034334

cc: @carsonfarmer

joewagner:
The reason this is happening is because the SDK is expecting the response from a call to `receipt` to have a top level `error` property, example below:
```
{
    "jsonrpc": "2.0",
    "id": 69,
    "result": {
        "ok": true,
        "error": "db query execution failed (code: SQLITE_NOT NULL constraint failed: leaderboard_5_439.id, msg: NOT NULL constraint failed: leaderboard_5_439.id)",
        "error_event_idx": 0
        "receipt": {
            "chain_id": 5,
            "txn_hash": "0xde3594957f955bfc52ce861ab9d0b0c222653e0febb5cb2c5ef2093e33d4cda8",
            "block_number": 7465202
        }
    }
}
```


However in the case that the transaction's SQL failed due to a table constraint violation the response looks like below:
```
{
    "jsonrpc": "2.0",
    "id": 69,
    "result": {
        "ok": true,
        "receipt": {
            "chain_id": 5,
            "txn_hash": "0xde3594957f955bfc52ce861ab9d0b0c222653e0febb5cb2c5ef2093e33d4cda8",
            "block_number": 7465202,
            "error": "db query execution failed (code: SQLITE_NOT NULL constraint failed: leaderboard_5_439.id, msg: NOT NULL constraint failed: leaderboard_5_439.id)",
            "error_event_idx": 0
        }
    }
}
```

We can update the SDK to find and throw both of these cases, or we could update the structure of the response the Validator is sending.
@jsign Do you any thoughts on what we should do here?

carsonfarmer:
According to the [JSON-RPC spec](https://www.jsonrpc.org/specification#response_object), a top level error key is required on error. So we should probably go with that response from the API. With that in mind, we are _already_ not returning spec compliant responses it seems. Because error should actually an object:

[...]

PR #122, opened 2022-08-25
Joe/fix 121

joewagner:
Fix for #121


carsonfarmer (approved):
Yes I think this is the right flow. Remind me, is skipConfirm true or false by default?

joewagner:
> Yes I think this is the right flow. Remind me, is skipConfirm true or false by default?

`skipConfirm` is false by default, which means by default we **do** wait for confirmation.  We could change that pretty easy if needed.

carsonfarmer:
Nope, that's perfect. Wait for confirmation by default is nice, let's keep that.

Issue #116, opened 2022-08-21
`fetch is not defined` error due to missing `fetch` import in `tableland-calls.ts`


dtbuchholz:
### Describe the bug
I'm trying to simply use Tableland in an ES module, but I _always_ get the error `fetch is not defined`, stemming from `tableland-calls.js` (using `@tableland/sdk@3.1.0`). I use `node v18.7.0` and `npm 8.15.0`. My `package.json` has a `type: module` defined, and I successfully `import fetch from "node-fetch"` and use `fetch` elsewhere in my code. I try replacing my `fetch` import with the following 1-liner in my code (as recommended per the JS SDK docs for CJS/`require`), but it has no impact:
```
const fetch = (...args) => import("node-fetch").then(({ default: fetch }) => fetch(...args))
```
The only way to stop this error in my application is (usually) by reverting to CJS over ESM and then using the 1-liner/polyfill, which is not ideal. Thus, I'm assuming the issue must originate from Tableland, not my code. 

After adding this 1-liner to `tableland-calls.js` in `dist`, the error is resolved. Upon digging into the SDK code a bit, it looks like the error stems from the fact `tableland-calls.ts` does not import `node-fetch`. If you import `node-fetch` into `tableland-calls.ts` and build, everything is fixed; no more `fetch is not defined` error.

**Disclaimer**: I know we use a modern version of Node, but I'm using Node v18 and still having this issue...so I'm not sure exactly why this issue is happening, to be honest. I would expect that I wouldn't have to import `node-fetch` with Node 18. Albeit, I'm not an expert on these intricacies. The fix described below resolves the error, and I've heard this problem pop up a few times from others, before. Not exactly sure if they were using CJS or ESM or what version of Node, though.

### To Reproduce
1. Create some Node.js ESM (using JS!) and install deps `node-fetch` as well as `@tableland/sdk`; call the file `index.js`.
2. Import these into your `index.js`:
```
import fetch from "node-fetch"
import { connect } from "@tableland/sdk"
```
3. Try to use the SDK's `read` query after connecting. You will get an error:
```
file:///Users/db/Tableland/projects/discord-table-bot/node_modules/@tableland/sdk/dist/esm/src/lib/tableland-calls.js:22
        const res = yield fetch(`${this.options.host}/rpc`, {
                    ^
ReferenceError: fetch is not defined
```
4. Try to fix it with the 1-liner _within_ `index.js`, and nothing is resolved.
[...]

PR #148, opened 2022-09-12
fix #129 use Buffer polyfil if not already global

### LINKED ISSUE #129: As a side effect of updating to siwe v2 the SDK now relies on an implementation of Buffer being available on the global scope
As a side effect of updating to siwe v2 the SDK now relies on an implementation of Buffer being available on the global scope.

More discussion here: https://github.com/tablelandnetwork/js-tableland/pull/112#discussion_r949049876

joewagner:
@carsonfarmer Here's a simple fix for #129.  Do we want to do this for people, or require them to setup their own solution?

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -1,4 +1,5 @@
 import { Signer, ethers } from "ethers";
+import BufferPolyfil from "buffer";
 import camelCase from "camelcase";
 import * as evm from "@tableland/evm/proxies.js";
 import {
@@ -27,9 +28,11 @@ const polyfills = {
 
     return buffer.toString("base64");
   },
+  Buffer: BufferPolyfil,
 };
 
 export const btoa = globalThis.btoa ?? polyfills.btoa;
+export const Buffer = globalThis.Buffer ?? polyfills.Buffer;
 
 export type NetworkName = "testnet" | "staging" | "custom";
 


carsonfarmer (approved):
This is great, thanks for the quick fix @joewagner. Let's get this in the next release. It is simple enough that there's no need to require folks to include it themselves. 

Honestly, we should all be moving away from Buffer in favour of UInt8Array, but I guess siwe missed the memo 😜

joewagner:
> This is great, thanks for the quick fix @joewagner. Let's get this in the next release. It is simple enough that there's no need to require folks to include it themselves.
> 
> Honestly, we should all be moving away from Buffer in favour of UInt8Array, but I guess siwe missed the memo 😜

Haha, apparently so! Merging now, so the next release will get this

dtbuchholz:
question -- does it make sense to do this for node-fetch, too? i.e., the issue would be to "use fetch polyfill if not already global"...which i suppose would just be https://github.com/tablelandnetwork/js-tableland/issues/116, again.

i noted in that linked issue that i'm going to check back in once i figure out machine issues. but, it's pretty obvious that people install the SDK and don't read the fine print around using < Node 18. it's probably the most common installation issue (e.g., another example [here](https://discord.com/channels/592843512312102924/902337528902860820/1019243446961176656)) and leads to the same questions in the Discord. 

alternatively, is there a way we could add some logging to implement the fetch polyfill when using < Node 18, like:
```
Please use Node 18.x.x or provide global access to node-fetch using the polyfill defined here: https://github.com/node-fetch/node-fetch#providing-global-access`?
```

Issue #33, opened 2022-04-07
Source map points to missing source files


joewagner:
The build of this code base that is published to npm contains `.map` files with reference to the src directory.  The src directory is not published to npm which means that frontend tooling like [Vite](https://vitejs.dev/) send out many Warnings anytime this package is used.

look in `dist/esm/src/main.js.map`, you will see the following JSON

```
{"version":3,"file":"main.js","sourceRoot":"","sources":["../../../src/main.ts"],"names":[],"mappings":"AAAA,OAAO,EAAE,OAAO,EAAE,MAAM,oBAAoB,CAAC;AAC7C,cAAc,iBAAiB,CAAC"}
```

Note that the "sources" property refers to `../../../src/main.ts`.  That file is not published to npm, hence the warnings from Vite

joewagner:
Looks like the "files" prop in package.json is used to exclude the `src/` directory

carsonfarmer:
Yup exactly, we can just add the src dir to files to fix this.

PR #158, opened 2022-09-15
enable new validator read response formatting

joewagner:
This will change the default response to calling the `read` method.  As such will need a major version bump, and clear messaging to current users.

### DIFF
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -69,8 +69,10 @@ export interface CreateOptions {
   timeout?: number;
 }
 
+export type ReadOutput = "table" | "objects";
+
 export interface ReadOptions {
-  output?: string;
+  output?: ReadOutput;
   unwrap?: boolean;
   extract?: boolean;
 }
--- a/src/lib/connection.ts
+++ b/src/lib/connection.ts
@@ -69,6 +69,12 @@ export interface CreateOptions {
   timeout?: number;
 }
 
+export interface ReadOptions {
+  output?: string;
+  unwrap?: boolean;
+  extract?: boolean;
+}
+
 export interface WriteOptions {
   skipConfirm?: boolean;
   rpcRelay?: boolean;
--- a/src/lib/tableland-calls.ts
+++ b/src/lib/tableland-calls.ts
@@ -5,6 +5,7 @@ import {
   WriteQueryResult,
   ReceiptResult,
   Connection,
+  ReadOptions,
 } from "./connection.js";
 import { list } from "./list.js";
 import { camelCaseKeys } from "./util.js";
@@ -89,6 +90,7 @@ async function GeneralizedRPC(
     | GetReceiptParams
     | SetControllerParams
     | LockControllerParams
+    | ReadOptions
 ) {
   return {
     jsonrpc: "2.0",
@@ -128,9 +130,14 @@ async function validateWriteQuery(
   return camelCaseKeys(json.result);
 }
 
-async function read(this: Connection, query: string): Promise<ReadQueryResult> {
+async function read(
+  this: Connection,
+  query: string,
+  options?: ReadOptions
+): Promise<ReadQueryResult> {
   const message = await GeneralizedRPC.call(this, "runReadQuery", {
     statement: query,
+    ...options,
   });
   const json = await SendCall.call(this, message);
 
--- a/src/lib/tableland-calls.ts
+++ b/src/lib/tableland-calls.ts
@@ -133,8 +133,9 @@ async function validateWriteQuery(
 async function read(
   this: Connection,
   query: string,
-  options?: ReadOptions
+  options: ReadOptions = {}
 ): Promise<ReadQueryResult> {
+  if (!options.output) options.output = "table";
   const message = await GeneralizedRPC.call(this, "runReadQuery", {
     statement: query,
     ...options,
@@ -206,31 +207,4 @@ async function setController(
   return camelCaseKeys(json.result.tx);
 }
 
-// NOTE: tablelandCalls.lockController is not being used because the
-//       Validator RPC endpoint has not been implemented yet.
-async function lockController(
-  this: Connection,
-  tableId: string
-): Promise<WriteQueryResult> {
-  const message = await GeneralizedRPC.call(this, "lockController", {
-    token_id: tableId,
-  });
-  if (!this.token) {
-    await this.siwe();
-  }
-
-  const json = await SendCall.call(this, message, this.token?.token);
-
-  return camelCaseKeys(json.result.tx);
-}
-
-export {
-  hash,
-  list,
-  receipt,
-  read,
-  validateWriteQuery,
-  write,
-  setController,
-  lockController,
-};
+export { hash, list, receipt, read, validateWriteQuery, write, setController };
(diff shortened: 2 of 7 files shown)


joewagner:
Not much to review here yet, but tagging reviews just for an early heads up

joewagner:
This PR is ready for review.  The SDK changes are fairly simple, but there are a few things I want to mention:

- We can't publish the SDK changes in this PR to npm until the Validator API changes in [#258](https://github.com/tablelandnetwork/go-tableland/pull/258) take affect.  When the changes in #258 take affect any apps using the current SDK version for reads will probably break.  Because we can't publish the changes in this PR until changes in #258 take affect, SDK users can't really avoid having their apps being broken for a period of time.  Basically, we are in a catch 22 situation. In the past we have handled this by having the Validator support both versions for some period of time, usually about a week.  Open question, is there anyway to serve both API versions?  Maybe there's an alternative way to avoid breaking existing apps?
- There are tests for these changes in the local-tableland [joe/npx-test branch](https://github.com/tablelandnetwork/local-tableland/tree/joe/npx-test/test), and there's a [wip PR #33](https://github.com/tablelandnetwork/local-tableland/pull/33) that should result in this module being able to add @tableland/local as a dev dependency and using it to run tests.  Doing that would mean running the tests requires docker.  That might be worth a bigger discussion outside this PR.
- last thing, I added some more mocked tests to increase coverage for the controller methods.  Wasn't necessary, and if they introduce any problems we could split them into a different PR.

dtbuchholz:
The breaking changes aspect is a pretty big one here. We have 2 hackathons running until the end of the month, and then the pilot program kicks off a day or two after these end. Thus, we should at least wait for that in between period before we push anything. And once we align on a release date, we can start amplifying this on Discord to make sure people are aware it's coming.

carsonfarmer:
Hmmm, these are good points about breaking changes. The only real change is the default response type (`objects` vs what is now `tabular`). What about if we actually just push a patch update to the SDK now that bridges the gap for us. We could have it send the new param options to the validator to use the current default. Will the remote silently ignore extra params, or will this cause an issue?

src/lib/connection.ts
```diff
@@ -69,6 +69,12 @@ export interface CreateOptions {
   timeout?: number;
 }
 
+export interface ReadOptions {
+  output?: string;
```
carsonfarmer, line 73: Can we have a string enum here so that there's no ambiguity about what the valid options are?

src/lib/tableland-calls.ts
```diff
@@ -199,6 +206,8 @@ async function setController(
   return camelCaseKeys(json.result.tx);
 }
 
+// NOTE: tablelandCalls.lockController is not being used because the
+//       Validator RPC endpoint has not been implemented yet.
```
carsonfarmer, line 210: I suspect it won't be created either.

carsonfarmer:
Based on this:

```
https https://testnet.tableland.network/rpc \
  jsonrpc=2.0 id=1 method=tableland_runReadQuery \
  params:='[{"statement":"SELECT * FROM rpc_counter_5_65", "output": "objects"}]'
```

It looks like extra params are ignored? Am I testing this correctly?

joewagner:
> What about if we actually just push a patch update to the SDK now that bridges the gap for us. We could have it send the new param options to the validator to use the current default. Will the remote silently ignore extra params, or will this cause an issue?

This could work! Sorry if I missed this when we were chatting.  
I don't know know if the extra param would cause any issues, but I can figure that out pretty quick.
Just to clairify, the idea is we put out a version 3.1.2 that internally sends the `output: "table"` param with all reads.  Then make sure everyone has time to update to 3.1.2 before we merge #258. Then stop sending the param when we release 4.0.0.
Is that right?

src/lib/tableland-calls.ts
joewagner: Should we remove this function altogether? If relaying was to happen putting this back would be simple

carsonfarmer: Let's hold off on any changes here until the discussion around APIs settles.

carsonfarmer:
> Just to clairify, the idea is we put out a version 3.1.2 that internally sends the `output: "table"` param with all reads.

Correct

> Then make sure everyone has time to update to 3.1.2 before we merge #258.

Do we need to though? If we publish 3.1.2 now, it'll "just work" with the current setup. And then, if we (silently) merge #258 on go-tableland, things will still "just work" for SDK users, right?

> Then stop sending the param when we release 4.0.0.

Totally, this would clearly indicate that there are new default response types, and that updates will need to happen on the client side. But for those that don't want to change yet, 3.1.2 should still work as expected, because it is already sending the `output: "objects"` param.

joewagner:
@carsonfarmer All sounds good!  Just pushed that change we worked through in the voice channel.  Double check me then I think we can merge and release this as a patch.

carsonfarmer (approved):
Awesome, these latest updates are killer. This will keep users from having to do anything right away, and set us up nicely for deploying the backend. The only sticking point (and mentioned in the thread) is that _older_ versions of the SDK will continue to work, but will _also_ start to behave just like the new 4.0.0 release once we get that out (i.e., responses will be objects rather than tables). Not much we can do about that without API versioning on the backend, which seems pre-mature at this stage.

src/lib/tableland-calls.ts
```diff
@@ -128,13 +130,19 @@ async function validateWriteQuery(
   return camelCaseKeys(json.result);
 }
 
-async function read(this: Connection, query: string): Promise<ReadQueryResult> {
+async function read(
+  this: Connection,
+  query: string,
+  options: ReadOptions = {}
+): Promise<ReadQueryResult> {
+  if (!options.output) options.output = "table";
```
asutula, line 138: Sorry I didn't get to reviewing this earlier. The default behavior is defined on the validator and is `objects`. I don't think we want to override that here.

asutula: Oh I now see this is to force the old behavior for existing users, sounds good, nice call!

src/lib/tableland-calls.ts
```diff
@@ -128,13 +130,19 @@ async function validateWriteQuery(
...
   const message = await GeneralizedRPC.call(this, "runReadQuery", {
     statement: query,
+    ...options,
   });
   const json = await SendCall.call(this, message);
 
-  return camelCaseKeys(json.result.data);
+  return json.result.data;
```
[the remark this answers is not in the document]
joewagner: Good callout,.`camelCaseKeys` is used elsewhere, and removing is worth considering.  The original idea was that in Java/typescript variable naming convention is more commonly lower camel case than snake case. That might have been a short sided idea, but we use this currently to ensure the method calls return the same properties regardless of if the relay is used or not.  
My opinion is that it makes sense to continue to convert to lower camel case as long as we use a relay. This is because ethers.js uses lower camel case, and because we have to do some kind of conversion we might as well go with the LCC since that seems to be more JS friendly.  The LCC vs SC is definitely just a preference, so I can definitely be talked out of this if it doesn't make sense to others.
@carsonfarmer has a TPD in Notion for better aligning of apis across everything which feels related to this. https://www.notion.so/textile/Tableland-public-APIs-across-platforms-are-inconsistent-33d1fbac20144d97b88ec889458f7f79

PR #194, opened 2022-10-07
fix #161 and fix a few more paths to dist files

### LINKED ISSUE #161: Expose ReadOptions on read method in Connection Interface.
Currently, we don't expose the ReadOption type on the read method in the public API. This isn't a problem per se, because we haven't used them in practice. But, we will soon, so we'll need to enable those optional params.

joewagner:
@carsonfarmer I found two more spots that where using the dist/esm/src/ path.
Also including a fix for #161 here

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
   },
   "type": "module",
   "main": "./dist/cjs/main.js",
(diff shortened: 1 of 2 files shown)


sanderpick (approved):
lgtm!

PR #202, opened 2022-10-17
Add Arbitrum mainnet fix #167

### LINKED ISSUE #167: Release SDK with Arbitrum support

joewagner:
Adds Arbitrum mainnet to list of tableland networks.
Notes:
 - Looks like we updated our eslint dependency and we have a lot of linting fixes all of a sudden.  I'll investigate later today
 - Need to ensure that evm-tableland proxies -> network file name is published and incorporated correctly with this pull request and publish.
cc: @carsonfarmer If you have time for a sync voice on getting this and evm published let me know.

### DIFF
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -101,11 +101,11 @@ export const SUPPORTED_CHAINS: Record<ChainName, SupportedChain> = {
     host: "https://testnet.tableland.network",
     rpcRelay: true,
   },
-  "arbitrum": {
+  arbitrum: {
     name: "arbitrum",
     phrase: "Arbitrum",
     chainId: 42161,
-    contract: evm.proxies["arbitrum"],
+    contract: evm.proxies.arbitrum,
     host: "https://testnet.tableland.network",
     rpcRelay: false,
   },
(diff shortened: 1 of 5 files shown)


src/lib/util.ts
```diff
@@ -40,6 +40,7 @@ export type ChainName =
   | "ethereum"
   | "optimism"
   | "polygon"
+  | "arbitrum"
```
sanderpick, line 43: is it possible to somehow pull these chain names from `evm.proxies`? we should take the opportunity to make evm.proxies just work so that there's no need to update this file when chains are added

joewagner: Yeah, that should be possible.  I'll set that up

carsonfarmer (approved):

src/lib/util.ts
carsonfarmer: We'll do that with a follow up PR.

joewagner: Just chatting with @carsonfarmer looks like this is more complicated than we thought.  If you're ok with it, can we tackle this in a bigger PR after a some planning?

sanderpick:
Sounds good!

src/lib/util.ts
sanderpick: sounds good! can you guys capture it in an issue so we can come back to it during planning?

PR #225, opened 2022-10-24
Fix missing issuedAt field when creating SiweMessage's

datadanne:
## Summary

Adds a missing `issuedAt` field when creating a `SiweMessage`. https://github.com/spruceid/siwe released a new version recently that made the issuedAt key required.


joewagner (approved):
@datadanne This is great!

PR #258, opened 2022-11-03
docs: update README

dtbuchholz:
## Summary

Devs often run into `node-fetch` issues and don't know what to do. The docs site tries to call this out early (not always effective...), but the README does not make mention of the requirement for Node 18+.

## Details

Adds context on Node 18 requirement and how to import `node-fetch` with <v18.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ npm i @tableland/sdk
 
 :warning: Please ensure you are using Node 18 (or newer).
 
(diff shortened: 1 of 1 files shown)


joewagner:
@dtbuchholz This is great.  Do we want to mention the possibility of using node 16 with the `--experimental-fetch` flag?

joewagner (approved):

dtbuchholz:
good idea! added.

PR #260, opened 2022-11-03
Remove node-fetch dev dependency

joewagner:
We had the node-fetch module as a dev dependency so that we could run end to end tests, but that's no longer required since we aren't running test with node v14.


sanderpick (approved):
nice to see that's possible!

PR #275, opened 2022-11-16
Add mainnet gateway

joewagner:


sanderpick:
nice! lg. tableland.network is functional now, so may as well get this in.

sanderpick (approved):

joewagner:
> nice! lg. tableland.network is functional now, so may as well get this in.

@carsonfarmer has a big change in the works. Before we merge this, I wanted to double check that it won't create a big conflict for him.

joewagner:
We will need this in version 3 of the sdk one way or another though.

carsonfarmer:
Naw no conflict here. Merge away!

carsonfarmer:
Actually, the set of big changes are a total refactor... so there are going to be conflicts no matter what. Don't hesitate to get things in that are needed.

awmuncy (approved):

jsign:
@joewagner, @carsonfarmer, I'm a bit confused by this PR so asking the question here.
I see that the SDK changed the gateway to point to the _future_ mainnet one.

Doesn't this mean that if we tag a release since this merge date, people will be using a gateway that they shouldn't? (i.e: it will be official at the first ship date)

But maybe you're planning to not create new tags until that happens, if that's the case this merged PR should be fine.
If that's the case, I see that the testnet one wasn't changed to `testnets.tableland.network` (at least in this PR). Was that a miss? Or maybe was/will be part of other PR?

joewagner:
@jsign Thanks for bringing this up.  I originally made this PR thinking it would sit until the ship date.  The comments above suggested that the new gateway was working and we should merge and publish this.  I double checked by doing a simple check of looking at https://tableland.network/schema/healthbot_1_1, but that might not be sufficient.
These changes have been published, so the new gateway is already being used by the SDK for mainnet connections.

jsign:
@joewagner, yeah, I think there was some confusion in that validator being used for other stuff too. For example, the Registry SC is targeting this validator too which was described as "planned" but was actually done by Sander.

Maybe we can leave things as they're. Since in mainnet we don't do transaction relaying, then we wouldn't have wallet problems of two validators having the same private key. 

In my ideal world, I'd like nobody to use this validator until we do the official ship (because if not, the "official ship" is somewhat false and is shipped today). But that's not entirely true already (not only by this SDK situation, but because of the TokenURI).

So.. let's simply leave things as they're (SDK mainnet targeting this new validator).
Let's just be aware of including that `testnets.tableland.network` for testnet chains for the next Ship as planned.

joewagner:
That makes sense.  I'll setup another issue and PR for the `testnets.tableland.network` change, and leave it unmerged until we ship. Thanks again for noting this. It seems kind of subtle, but in reality these kind of things will cause problems if they aren't tended to.

PR #330, opened 2022-12-20
feat: SDK refactor to provide D1Database compatible API plus helper functions/classes for working with validators and registries

### LINKED ISSUE #146: Create SDK variant that provides the D1Database Interface
A particularly good example is https://socket.dev/npm/package/@cloudflare/d1. Additionally, supporting or adopting this type of API means all d1 users are also Tableland users.

### LINKED ISSUE #205: Feat: Pull all chain and network information (and types) from @tableland/evm dependency.
If done properly, this would mean that an update to the underlying lib would automatically update all supported chains in @tableland/sdk and @tableland/cli. It also reduces repeated information, thereby reducing the likelihood of missed or incorrect chain information.

### LINKED ISSUE #232: The JS SDK implements the new versioned HTTP API

### LINKED ISSUE #234: Pull wasm-sqlparser directly into the SDK

### LINKED ISSUE #235: Drop support for relaying transactions via the gateway

### LINKED ISSUE #237: The SDK allows users to reducing code by supporting a small subset of common SQL template characters

### LINKED ISSUE #238: The SDK contains new tests for SQL templating

### LINKED ISSUE #241: Finalize approach to ORM support (implement a simple ORM layer or leverage existing tooling)

### LINKED ISSUE #243: Implement ORM support for the SDK by supporting a single or multiple ORM frameworks

### LINKED ISSUE #244: Implement new tests to support ORM usage patterns

### LINKED ISSUE #262: Make error message meaningful if this lib cannot get a signer
**Describe the bug**
If no signer is provided to the connect method, this lib will assume the environment is a browser and try to get the signer from the global `ethereum` object.  If `globalThis.ethereum` doesn't exist the error will be "Cannot read properties of undefined (reading 'request')".
This error is not very useful.  We should add a check [here](https://github.com/tablelandnetwork/js-tableland/blob/main/src/lib/util.ts#L18) that does something like:

```
if (!globalThis.ethereum) {
  throw new Error("Tableland: no signer provided and cannot get a signer from global scope");
}
```
 

**To Reproduce**

1. connect to tableland without providing a signer in an enviroment without globalThis.ethereum

**Expected behavior**
A useful error should be thrown

### LINKED ISSUE #271: Why are CreateTableReceipt fields optional?
**Is your feature request related to a problem? Please describe.**
Yes, the `CreateTableReceipt` fields `name`, `prefix`, and `tableId` are optional (`?`), but I can't think of why they should be. Presumably, the SDK will detect any receipt `error` and throw it, otherwise return the resulting data to the SDK user, which should always include those properties.

**Describe the solution you'd like**
They should not be optional.

**Describe alternatives you've considered**
Dealing with the optionals which is a pain when it's not necessary.

**Additional context**
None

### LINKED ISSUE #298: Ensure we  block reads that attempt to join tables between test- and main-nets
The current version of the SDK requires a user to be connected to a given chain, be it a testnet or mainnet. However, it doesn't strictly enforce that queries must only target that chain. We should catch this potential error on the client _before_ it is submitted to validators to prevent unnecessary API calls. This requires work from @tableland/sqlparser (https://github.com/tablelandnetwork/wasm-sqlparser/issues/54) to be resolved.

Note that this work will already be included in the refactored SDK.

### LINKED ISSUE #309: Update testnet chains connection objects to point to new testnets gateway
The upcoming ship of the new gateway will include a change in the url from https://testnet.tableland.network to https://testnets.tableland.network
We need to make that change in util.ts and ensure that publishing the change happens in coordination with the ship

carsonfarmer:
This is a major refactor (re-write) of the TypeScript/JavaScript SDK. It includes many new features, fixes, tests, and design changes. This summary and the subsequent comments should help you review the changes, which are many. If you want to jump on a call with @carsonfarmer to go over changes live, then please don't hesitate to ask.

## Summary

This PR exposes three core APIs to user-land. The "primary" API, is a [D1 client compatible database API](https://developers.cloudflare.com/d1/platform/client-api/). It is a small and simple API that will feel very familiar to web2 database users. It includes the concept of prepared statements, SQL parameter binding, execution and query modes, and more. It is similar to APIs such as [better-sqlite3](https://github.com/WiseLibs/better-sqlite3), [D1](https://blog.cloudflare.com/introducing-d1/), and more.

There are two additional "secondary" APIs. These are specific to Tableland. The first is the Validator API. This wraps the set of core APIs provided by Tableland Validators. The spec for this API comes from our [spec documentation](https://github.com/tablelandnetwork/docs/tree/main/specs/validator), and the underlying client is auto-generated from that. The next "secondary" API is the Registry API. This wraps interactions with the TablelandTables smart contract. It enables on-chain access control, sending create and mutate queries, and more.

In general, users will focus on the Database API, and only when fine-grained control, or specific use-cases require it, will they need to leverage the two secondary APIs. Note that there are also a ton of helper functions included to help with things like testing, accessing Tableland network defaults, and interacting with wallets, chains, and other Tabeland features.

## Details

Details are in the PR comments for the most part. There are a lot of them! Dig in and let me know if you need clarification on anything.

Fixes #146
Fixes #205
Fixes #232
Fixes #234
Fixes #235
Fixes #237
Fixes #238
Fixes #241
Fixes #243
Fixes #244
Fixes #262
Fixes #271
Fixes #298
Fixes #309

Partially addresses #244

## How it was tested

I've added full test coverage. Every line is either fully tested, or where testing fall-back scenarios seems overkill, I've added coverage exceptions (this is only a small number of cases). In total, there are 128 passing tests. These should all run in CI in about 3 minutes.

```text
----------------------|---------|----------|---------|---------
File                  | % Stmts | % Branch | % Funcs | % Lines 
----------------------|---------|----------|---------|---------
All files             |   98.91 |    96.39 |   98.44 |   98.91
 src                  |     100 |      100 |     100 |     100 
  database.ts         |     100 |      100 |     100 |     100 
  index.ts            |     100 |      100 |     100 |     100 
  lowlevel.ts         |     100 |      100 |     100 |     100 
  statement.ts        |     100 |      100 |     100 |     100 
 src/helpers          |     100 |      100 |     100 |     100 
  await.ts            |     100 |      100 |     100 |     100 
  binding.ts          |     100 |      100 |     100 |     100 
  chains.ts           |     100 |      100 |     100 |     100 
  config.ts           |     100 |      100 |     100 |     100 
  ethers.ts           |     100 |      100 |     100 |     100 
  index.ts            |     100 |      100 |     100 |     100 
  parser.ts           |     100 |      100 |     100 |     100 
  utils.ts            |     100 |      100 |     100 |     100  
 src/registry         |     100 |      100 |     100 |     100 
  contract.ts         |     100 |      100 |     100 |     100 
  controller.ts       |     100 |      100 |     100 |     100 
  create.ts           |     100 |      100 |     100 |     100 
  index.ts            |     100 |      100 |     100 |     100 
  run.ts              |     100 |      100 |     100 |     100 
  tables.ts           |     100 |      100 |     100 |     100 
  transfer.ts         |     100 |      100 |     100 |     100 
  utils.ts            |     100 |      100 |     100 |     100 
 src/validator        |     100 |      100 |     100 |     100 
  errors.ts           |     100 |      100 |     100 |     100 
  health.ts           |     100 |      100 |     100 |     100 
  index.ts            |     100 |      100 |     100 |     100 
  query.ts            |     100 |      100 |     100 |     100 
  receipt.ts          |     100 |      100 |     100 |     100 
  tables.ts           |     100 |      100 |     100 |     100 
  version.ts          |     100 |      100 |     100 |     100 
 src/validator/client |   94.56 |    76.78 |   92.59 |   94.56 
  fetcher.ts          |    90.8 |    75.92 |      92 |    90.8 
  index.ts            |     100 |      100 |     100 |     100                                   
  types.ts            |     100 |      100 |     100 |     100 
----------------------|---------|----------|---------|---------
```

The tests provide lots of really nice usage examples, and even a few anticipated uses of third party libraries. I'd like to also leverage more docs testing, but we have full 100% test coverage for now, so that is a very nice place to start from.

## Checklist:


src/database.ts
```diff
@@ -0,0 +1,100 @@
...
+  constructor(config: Config & Partial<AutoWaitConfig> = {}) {
+    this.config = config;
+  }
+
+  static async readOnly(chainNameOrId: ChainName | number): Promise<Database> {
+    const baseUrl = getBaseUrl(chainNameOrId);
+    return new Database({ baseUrl });
+  }
+
+  static async forSigner(signer: Signer): Promise<Database> {
```
carsonfarmer, line 26: This will determine which baseUrl to use given the chain that this signer is connected to. These two static methods are probably the best ways for devs to construct a new Database instance.

src/database.ts
```diff
@@ -0,0 +1,100 @@
...
+    const baseUrl = await extractBaseUrl({ signer });
+    return new Database({ signer, baseUrl });
+  }
+
+  prepare<T = D>(sql: string): Statement<T> {
```
carsonfarmer, line 31: The primary method for this Database interface. I'll discuss more in the Statement class what this means, but TL;DR is that we are currently "faking" prepared statements on the client-side to a) mimic the D1 API, and b) prepare for prepared statements (haha) when they actually land in Tableland.

src/database.ts
```diff
@@ -0,0 +1,100 @@
...
+    return new Statement<T>(this.config, sql);
+  }
+
+  /**
+   * @param statements
+   * @returns
+   */
+  async batch<T = D>(
```
carsonfarmer, line 39: On Tabeland, batching statements together requires some finesse. This is the first pass at this. Ideally, we can do more "smart" splitting to handle things like mixing creates, mutations, and reads. Interestingly, the D1 API treats batches as a kinda unique thing, where multiple reads would lead to multiple responses etc.

Note here that writes are done as a single "transaction" (in the smart contract sense), and reads are done "concurrently". This means that writes _have_ to follow the rules of our validator in terms of grouping. You can't mix creates with anything else, you can't mix reads and writes, but you can mix revoke/grant with writes.

src/database.ts
```diff
@@ -0,0 +1,100 @@
...
+    } catch (cause: any) {
+      if (cause.message === "ALL_ERROR") {
+        throw new Error("BATCH_ERROR", { cause: cause.cause });
+      }
+      throw new Error("BATCH_ERROR", { cause });
+    }
+  }
+
+  // TODO: Note that this appears to be the wrong return type in practice.
+  async exec<T = D>(
```
carsonfarmer, line 77: So, this todo refers to the fact that, while the D1 interface says one thing... their public implementation appears to do another. I'm going by the interface because that's what folks'll be expecting. We might have to update once they are out of beta and things are more aligned?

carsonfarmer: Really, this is kinda a lame API and even the D1 docs say to avoid using it for the most part.

src/database.ts
```diff
@@ -0,0 +1,100 @@
...
+      return result;
+    } catch (cause: any) {
+      if (cause.message === "RUN_ERROR") {
+        throw new Error("EXEC_ERROR", { cause: cause.cause });
+      }
+      throw new Error("EXEC_ERROR", { cause });
+    }
+  }
+
+  async dump(_opts: Signal = {}): Promise<ArrayBuffer> {
```
carsonfarmer, line 97: A nice future feature will be to allow a user to dump a set of tables to a local sqlite database. That might require additional deps, so we're not doing it for now. Lots of other tools also don't support this, so I think that's ok.

src/database.ts
```diff
@@ -0,0 +1,100 @@
...
+    throw new Error("DUMP_ERROR", { cause: new Error("not implemented yet") });
```
carsonfarmer, line 98: You might notice these error codes with cause. This cause property is a new Error feature. While I like this pattern of including a cause a lot... it is really annoying when dealing with errors, because the printed error message is just "DUMP_ERROR". So to match the D1 API I'm using this, but it kinda sucks in practice IMO.

src/helpers/await.ts
```diff
@@ -0,0 +1,62 @@
+export type Awaitable<T> = T | PromiseLike<T>;
```
carsonfarmer, line 1: You can mostly skip this file. This is a series of helper interfaces and functions for dealing with promises.

src/helpers/await.ts
```diff
@@ -0,0 +1,62 @@
...
+    const controller = new AbortController();
+    abortSignal = controller.signal;
+    setTimeout(controller.abort.bind(controller), maxTimeout);
+  } else {
+    abortSignal = signal;
+  }
+  return abortSignal;
+}
+
+export async function getAsyncPoller<T = unknown>(
```
carsonfarmer, line 39: This is really the only interesting function. It simply wraps a polling pattern, but rather than a timeout of some sort, it exposes the AbortController API so that callers have more control over when they want to abort. Having said that, the default behavior is simply to create an abort controller that will abort after 10 seconds.

src/helpers/binding.ts
```diff
@@ -0,0 +1,163 @@
+const placeholderRegExp =
```
carsonfarmer, line 1: While you can probably just skip this for the purposes of this review... there is a lot of pretty cool magic in here. I mean, look at that regex! But truly, this module simply parses SQL statements, and does its best to do client-side binding of parameters. It was a lot of work, so please at least appreciate it!

Soonish, this can be dropped in favor of real prepared statements. Though quite a bit of the auto-encoding stuff will still be relevant. For instance, this does its best to encode things like floats, numbers, dates, and binary data into the right format for our SQL spec. For now, I allow floats to be passed through, but this will be checked by our parser so it should be safe.

src/helpers/binding.ts
```diff
@@ -0,0 +1,163 @@
...
+  /["'`](?:(?<=")[^"\\]*(?:\\.[^"\\]*)*"|(?<=')[^'\\]*(?:\\.[^'\\]*)*'|[^`]*`)|(\?\d*|[:@$][a-zA-Z_]\w+)/gmu;
+
+function isPlainObject(obj: any): obj is Record<string, BaseType> {
```
carsonfarmer, line 4: It is almost 2023, and it is still hard to know if something is a plan ol' javascript object or not...

src/helpers/binding.ts
```diff
@@ -0,0 +1,163 @@
...
+
+function bindNull(_param: undefined | null): string {
+  return "NULL";
+}
+
+function bindToString(param: any): string {
+  return bindString(String(param));
+}
+
+interface SQL {
```
carsonfarmer, line 43: A little bit of a magic thing if someone wants to stick a custom object into a table. This will likely be useful for things like auto-pinning data to IPFS and including the hash and things like that. This way, we have a single interface that the lib knows about to do all sorts of custom things.

src/helpers/binding.ts
```diff
@@ -0,0 +1,163 @@
...
+        /* c8 ignore next 3 */
+      } else {
+        return bindToString(param);
+      }
+    default:
+      return bindToString(param);
+  }
+}
+
+export type BaseType =
```
carsonfarmer, line 86: I actually define the types we support here. This _could_ probably be left as any, but this typed version is nice because it'll catch weird things for us. If using from JavaScript, folks can get around this, so it isn't really saving us from runtime weirdness.

src/helpers/binding.ts
```diff
@@ -0,0 +1,163 @@
...
+        return { anon: [...anon, v], named };
+      }
+    },
+    initialValue
+  );
+  return result;
+}
+
+export function bindValues(sql: string, parameters?: Parameters): string {
+  // https://sqlite.org/forum/forumpost/4350e973ad
```
carsonfarmer, line 122: I learned a lot about binding values and how SQLite handles all its parameter values in SQL strings.

src/helpers/chains.ts
```diff
@@ -0,0 +1,107 @@
+import {
```
carsonfarmer, line 1: The source of truth now for all supported chains stuff is @tableland/evm... this module simple takes that information, extracts what it needs, and exposes this stuff via helper functions. So it is really easy to determine which validator gateway to target based on a given query, or signer connection. You'll notice that we don't actually define _any_ chain specific state here at all ourselves. It all derives from @tableland/evm.

src/helpers/chains.ts
```diff
@@ -0,0 +1,107 @@
...
+  ChainInfo
+>;
+
+// Not exported
+const supportedChainsById = Object.fromEntries(
+  Object.values(supportedChains).map((v) => [v.chainId, v])
+);
+
+/**
+ * Get the default chain information for a given chain name.
```
carsonfarmer, line 50: All the getters are here, and they'll take a number or chain name. The chain names should match with ethersjs once the PR in @tableland/evm is published and we pull that in here.

src/helpers/chains.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Get the default host uri for a given chain name.
+ * @param chainNameOrId The requested chain name.
+ * @returns A string representing the default host uri for a given chain.
+ */
+export function getBaseUrl(chainNameOrId: ChainName | number): string {
+  return getChainInfo(chainNameOrId).baseUrl;
+}
+
+export function overrideDefaults(
```
carsonfarmer, line 92: This is a handy feature for testing, but could work for other folks as well. It allows one to override _any_ aspect of our supportedChains settings. So you can override the contract address for a given chain (including localhost as we do in tests), you can override the target validator base url if you want to use a custom one, etc

src/helpers/connection.ts
```diff
@@ -0,0 +1,45 @@
+import { type ChainName, getBaseUrl } from "./chains.js";
```
carsonfarmer, line 1: Actually, this file should be renamed to config. But this contains the limited bit of information the SDK needs to know to configure connections and behavior. There are only 3 settings, and each one is only relevant in certain contexts, so they are also their own interfaces. This allows us to be pretty lazy about checks. For instance, if you are only ever going to _query_ tableland, then you don't need any signer information, and vice versa.

src/helpers/connection.ts
```diff
@@ -0,0 +1,45 @@
...
+  signer: Signer;
+}
+
+export interface AutoWaitConfig {
+  autoWait: boolean;
+}
+
+export type Config = Partial<ReadConfig & SignerConfig>;
+
+export async function extractBaseUrl(
```
carsonfarmer, line 18: This function does its best to figure out where to route queries based on defaults, if you have a signer connected to a chain, if you have a fallback chain you want to use, etc.

src/helpers/ethers.ts
```diff
@@ -0,0 +1,100 @@
+import {
```
carsonfarmer, line 1: As far as possible, anything that needs to leverage the ethers dep should happen in this file. As you'll see it isn't a ton of stuff. There are some helpers, some interfaces, and several exports.

src/helpers/ethers.ts
```diff
@@ -0,0 +1,100 @@
...
+  // eslint-disable-next-line no-var
+  var ethereum: ExternalProvider | undefined;
+}
+
+/**
+ * Request a set of opinionated overrides to be used when calling the Tableland contract.
+ * @param signer A valid web3 provider/signer.
+ * @returns A promise that resolves to an object with overrides.
+ */
+export async function getOverrides({
```
carsonfarmer, line 25: Polygon annoyance that I pulled in from our existing SDK.

src/helpers/ethers.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+  return opts;
+}
+
+export type RegistryReceipt = Required<
+  Omit<TransactionReceipt, "error" | "errorEventIdx">
+>;
+
+export async function getContractReceipt(
```
carsonfarmer, line 46: Pulls out events from our registry contract.

src/helpers/ethers.ts
```diff
@@ -0,0 +1,100 @@
...
+  return { tableId, transactionHash, blockNumber, chainId };
+}
+
+/**
+ * Request a signer object from the global ethereum object.
+ * @param external A valid external provider. Defaults to `globalThis.ethereum` if not provided.
+ * @returns A promise that resolves to a valid web3 provider/signer
+ * @throws If no global ethereum object is available.
+ */
+export async function getSigner(external?: ExternalProvider): Promise<Signer> {
```
carsonfarmer, line 79: This will attempt to pull a signer from the global injected provider if it can't get one from anywhere else. This works for browsers, and fails on node. The errors should be helpful in figuring out what went wrong. This is a last ditch effort, and shouldn't really happen unless the user hasn't provided any connection information. Having said that, it is also a handy helper function that users might _want_ to use.

src/helpers/index.ts
```diff
@@ -0,0 +1,40 @@
+export {
```
carsonfarmer, line 1: Just defines the main public "helpers" API.

src/helpers/parser.ts
```diff
@@ -0,0 +1,53 @@
+import {
```
carsonfarmer, line 1: Wraps our parser functions and makes it a bit easier to use. This essentially lazily loads the wasm bundle the first time you use it. It also has a nice function for validating table names and checking that you aren't trying to mix and match tables across testnets and mainnets.

src/helpers/utils.ts
```diff
@@ -0,0 +1,9 @@
+export type PartialRequired<T, S extends keyof T> = Omit<Required<T>, S> &
```
carsonfarmer, line 1: I tried so hard to _not_ have any utils files... but alas, these are handy functions to have.

src/lowlevel.ts
```diff
@@ -0,0 +1,109 @@
+import {
```
carsonfarmer, line 1: These are the lower level APIs that are used to access the Tableland network. They aren't part of the public APIs, and are used by the Database and Statement classes to call out to Tableland. They essentially wrap the validator and registry APIs in the nice behavior that a Database or Statement expects.

src/lowlevel.ts
```diff
@@ -0,0 +1,109 @@
...
+  wrapTransaction,
+} from "./registry/utils.js";
+import {
+  type ObjectsFormat,
+  type ValueOf,
+  getQuery,
+} from "./validator/query.js";
+import { ApiError } from "./validator/index.js";
+
+export async function exec(
```
carsonfarmer, line 22: Mutating functions go in here. We break things into prepare X and then run X and then wrap X.

src/lowlevel.ts
```diff
@@ -0,0 +1,109 @@
...
+      const tx = await runSQL(_config, prepared);
+      return await wrapTransaction(_config, prefix, tx);
+    }
+    /* c8 ignore next 2 */
+    default:
+      throw new Error("invalid statement type: read");
+  }
+}
+
+function catchNotFound(err: unknown): [] {
```
carsonfarmer, line 50: Workaround for this "discussion"/issue: https://github.com/tablelandnetwork/docs/pull/53#discussion_r1051388080

src/lowlevel.ts
```diff
@@ -0,0 +1,109 @@
...
+    return [];
+  }
+  throw err;
+}
+
+export async function queryRaw<T = unknown>(
+  config: ReadConfig,
+  statement: string,
+  opts: Signal = {}
+): Promise<Array<ValueOf<T>>> {
```
carsonfarmer, line 61: Note the amazing typed responses we have here. This type is smart enough to _know_ that a raw response will basically unwrap a type, and just return its "values". Cool!

src/lowlevel.ts
```diff
@@ -0,0 +1,109 @@
...
+  statement: string,
+  opts: Signal = {}
+): Promise<T | null> {
+  const response = await queryAll<T>(config, statement, opts).catch(
+    catchNotFound
+  );
+  return response.shift() ?? null;
+}
+
+export function extractColumn<T = unknown, K extends keyof T = keyof T>(
```
carsonfarmer, line 90: Typed value extraction. Kinda cool here, the type can check if it is an array of values, an array of arrays, or whatever and return the right typed single value.

src/registry/contract.ts
```diff
@@ -0,0 +1,63 @@
+import {
```
carsonfarmer, line 1: Just abstracting away some things for interacting with our registry smart contract.

src/registry/controller.ts
```diff
@@ -0,0 +1,44 @@
+import { type SignerConfig } from "../helpers/connection.js";
```
carsonfarmer, line 1: Controller based method on the registry class. This pattern is used throughout the SDK. We have the core functions, that take a config, and then these are pulling into the exposed classes.

carsonfarmer: My goal with all of this is to hide all the glue code into helpers, making the actual work (like this) much easier to grok.

src/registry/create.ts
```diff
@@ -0,0 +1,59 @@
+import { type SignerConfig } from "../helpers/connection.js";
+import { type ContractTransaction } from "../helpers/ethers.js";
+import { validateTableName } from "../helpers/parser.js";
+import { getContractAndOverrides } from "./contract.js";
+
+// Match _anything_ between create table and schema portion of create statement
```
carsonfarmer, line 6: Ok, this is kinda crazy... but this is how I make it possible to extract a plain table name from something like:

```sql
CREATE TABLE my_table (...);
```
and combine it with the chainId information to create a new table name for a create table statement like this: `my_table_31337`. I ended up on a two part regex pattern, where we first take _anything_ that might be a name, and then parse it out as best as possible, dealing with escaping and things like that. It'll fail if you try to do anything particularly fancy with your table name. But things like [my_table], or `my_table` etc that many ORMs use should be safe.

src/registry/create.ts
```diff
@@ -0,0 +1,59 @@
...
+const secondSearch =
+  /(?<=^")(.*?)(?="$)(?<=^\[)(.*?)(?=\]$)|(?<=^')(.*?)(?='$)|(?<=^`)(.*?)(?=`$)|(^[^[`'].*?[^\]'`]$)/;
+
+export interface PrepareParams {
+  statement: string;
+  chainId: number;
+  first: string;
+}
+
+export async function prepareCreateTable({
```
carsonfarmer, line 19: prepare and do and wrap, same pattern here as elsewhere.

src/registry/create.ts
```diff
@@ -0,0 +1,59 @@
...
+  );
+  return { statement: stmt, chainId, prefix };
+}
+
+export interface CreateTableParams {
+  statement: string;
+  chainId: number;
+}
+
+export async function createTable(
```
carsonfarmer, line 49: the actual smart contract call

src/registry/index.ts
```diff
@@ -0,0 +1,131 @@
+import { type SignerConfig } from "../helpers/connection.js";
```
carsonfarmer, line 1: Public interface of the registry class. Most folks won't even use this, but if they want finer grained control over how they interact with the tableland network, they have some exposed methods here. We can easily expose more as needed.

src/registry/index.ts
```diff
@@ -0,0 +1,131 @@
...
+  type CreateTableParams,
+  type RunSQLParams,
+  type TransferParams,
+  type SetParams,
+};
+
+export class Registry {
+  constructor(readonly config: SignerConfig) {}
+
+  async listTables(owner?: string): Promise<TableIdentifier[]> {
```
carsonfarmer, line 33: This is missing any prefix information. I've tested out usage patterns based on this, and I think for the most part, this is fine. But we'll want to test that out in real life to be sure.

src/registry/run.ts
```diff
@@ -0,0 +1,40 @@
...
+  assertChainId,
+} from "./contract.js";
+
+export interface PrepareParams {
+  statement: string;
+  chainId: number;
+  first: string;
+}
+
+export async function prepareRunSQL({
```
carsonfarmer, line 16: Same pattern as with create. We don't try to parser or catch mistakes here at all (the tests also mention this). That is what the higher level APIs are for!

src/registry/utils.ts
```diff
@@ -0,0 +1,97 @@
+import {
```
carsonfarmer, line 1: Another utils file. Uhg. But this is the type of thing I want to "hide" from the main implementations, so here it is... in utils.ts...

src/statement.ts
```diff
@@ -0,0 +1,240 @@
+import {
```
carsonfarmer, line 1: GO HERE NEXT: This is actually the workhorse API of the entire thing now. This is where you "just write SQL". It is a really great API (not mine) and I find it really nice to work with. It took some deep thinking to make it fit a web3 world where we have to wait for transactions and things, but in the end, I think it is really intuitive! I hope you like it!

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+  queryAll,
+  queryFirst,
+  queryRaw,
+  exec,
+} from "./lowlevel.js";
+
+export { type ValuesType, type Parameters, type ValueOf, type BaseType };
+
+export class Statement<S = unknown> {
+  private readonly config: Config & Partial<AutoWaitConfig>;
```
carsonfarmer, line 32: Really, everything should be hidden. I'm using the "typescript" private keyword, even though I'd rather use the ES private pattern of using # prefixes. I might still do that, but apparently that breaks things in hardhat? cc @asutula

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+  bind<T = S>(...values: ValuesType[]): Statement<T> {
+    const parameters = getParameters(...values);
+    return new Statement(this.config, this.sql, parameters);
+  }
+
+  /**
+   * Resolve a bound statement to a SQL string.
+   * @returns A valid SQL string.
+   */
+  toString(): string {
```
carsonfarmer, line 62: This is _not_ part of the D1 spec, we use it a lot, and I don't like it. Once we have real prepared statements, it'll go away!

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+    try {
+      return bindValues(this.sql, this.parameters);
+    } catch (cause) {
+      throw new Error("BIND_ERROR", { cause });
+    }
+  }
+
+  toObject(): { sql: string; parameters?: Parameters } {
```
carsonfarmer, line 70: So you can see what your bound statement "looks like".

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+    return {
+      sql: this.sql,
+      parameters: this.parameters,
+    };
+  }
+
+  async #parseAndExtract(): Promise<ExtractedStatement> {
```
carsonfarmer, line 77: real private method... will this break hardhat? maybe?

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+    // Stick with original if a create statement, otherwise, use the parsed version
+    // This is because the parser injects keywords that are not spec compliant
+    // See https://github.com/tablelandnetwork/go-sqlparser/issues/41
+    const sql =
+      type === "create" ? statementWithBindings : statements.join(";");
+    return { type, sql, tables };
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async all<T = S, K extends keyof T = keyof T>(
```
carsonfarmer, line 89: The primary API. See how nicely typed it is?!

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+    colName?: K,
+    opts: Signal = {}
+  ): Promise<Result<T | T[K]>> {
+    try {
+      const start = performance.now();
+      const { sql, type, tables } = await this.#parseAndExtract();
+      switch (type) {
+        case "read": {
+          const config = await extractReadonly(this.config, {
```
carsonfarmer, line 106: Lazily fetch baseUrl information.

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+          if (colName != null) {
+            return wrapResult(
+              extractColumn(results, colName),
+              performance.now() - start
+            );
+          }
+          return wrapResult(results, performance.now() - start);
+        }
+        default: {
+          let receipt = await exec(this.config, { type, sql, tables });
```
carsonfarmer, line 120: all mutating queries happen here

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+          if (this.config.autoWait ?? false) {
```
carsonfarmer, line 121: Since D1 doesn't have the concept of deferred execution or transactions like we do, the downstream ORMs and things will return right after the smart contract call is made. If you turn on autoWait, they'll actually only resolve once the validator resolves the transaction. This is nice because it makes the API feel like a web2 API. But you can also turn this off and explicitly wait for things yourself, or not wait at all.

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+        cause.message.includes("column not found")
+      ) {
+        throw new Error("COLUMN_NOTFOUND", { cause });
+      }
+      throw new Error("ALL_ERROR", { cause });
+    }
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async first<T = S, K extends keyof T = keyof T>(): Promise<T>;
```
carsonfarmer, line 141: Not much special here, basically the same as all.

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+        cause instanceof Error &&
+        cause.message.includes("column not found")
+      ) {
+        throw new Error("COLUMN_NOTFOUND", { cause });
+      }
+      throw new Error("FIRST_ERROR", { cause });
+    }
+  }
+
+  async run(opts: Signal = {}): Promise<Result<never>> {
```
carsonfarmer, line 189: Essentially, only return metadata.

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+          }
+          return wrapResult(receipt, performance.now() - start);
+        }
+      }
+    } catch (cause) {
+      throw new Error("RUN_ERROR", { cause });
+    }
+  }
+
+  async raw<T = S>(opts: Signal = {}): Promise<Array<ValueOf<T>>> {
```
carsonfarmer, line 216: same as "all", but only returns raw rows (which uses format=table) in the validator api call.

src/validator/client/LICENCE
```diff
@@ -0,0 +1,22 @@
+MIT License
```
carsonfarmer, line 1: skip this file

src/validator/client/README.md
```diff
@@ -0,0 +1,185 @@
+> This is a local copy of openapi-typescript-fetch
```
carsonfarmer, line 1: skip this whole sub-module if you want. I stole this because it has some build issues, but the typescript code is perfectly good. it provides a really nice, typed fetch interface that works extremely well with our auto-build client from our openapi spec.

src/validator/client/fetcher.ts
```diff
@@ -0,0 +1,272 @@
+/* eslint-disable @typescript-eslint/consistent-type-assertions */
```
carsonfarmer, line 1: Ok, stopping here with comments for now because I have to run. More coming after I'm back!

carsonfarmer (commented):
First pass of a review helper pass. More to come later!

src/database.ts
```diff
@@ -0,0 +1,100 @@
...
+import { Statement } from "./statement.js";
+
+export class Database<D = unknown> {
+  readonly config: Config & Partial<AutoWaitConfig>;
+
+  constructor(config: Config & Partial<AutoWaitConfig> = {}) {
+    this.config = config;
+  }
+
+  static async readOnly(chainNameOrId: ChainName | number): Promise<Database> {
```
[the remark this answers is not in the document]
joewagner: Should we publish that evm commit under the next tag?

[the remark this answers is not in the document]
carsonfarmer: yeh some pre-release tag is fine!

test/statement.test.ts
```diff
@@ -0,0 +1,532 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
carsonfarmer, line 1: Like in the source, this tests is the workhorse test file of the suite. It is responsible for most of the coverage across the SDK. We have multiple tests for edge cases and specific expected behavior, as well as regression style tests pulled from past SDK experience.

test/statement.test.ts
```diff
@@ -0,0 +1,532 @@
...
+    const sql = "CREATE TABLE test (counter integer);";
+    const stmt = db.prepare(sql);
+    strictEqual(stmt.toString(), sql);
+    deepStrictEqual(stmt, new Statement(db.config, sql));
+  });
+
+  describe(".run()", function () {
+    let tableName: string;
+    this.beforeAll(async function () {
+      this.timeout("10s");
```
carsonfarmer, line 45: Most of these beforeAll hooks have a 10s timeout because they involve waiting for a create and insert to complete.

test/statement.test.ts
```diff
@@ -0,0 +1,532 @@
...
+    });
+
+    test("when a runtime error throws during a secondary wait", async function () {
+      const { meta } = await db
+        .prepare(
+          `INSERT INTO ${tableName} (id, counter, info) VALUES (1, ?1, ?2)`
+        )
+        .bind(5, "Bobby")
+        .all();
+      // TODO: Are we sure we don't want to use the cause pattern for these ones?
```
carsonfarmer, line 279: This is an open question. I honestly _don't_ like it per se, unless the message includes the cause message as well.

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
carsonfarmer, line 1: This is pretty cool. I like having this here because when we update these deps, we can make sure our lib is still compatible. Also, a great example of how to use these others libs with Tableland.

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
...
+  contractAddress: "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512",
+});
+
+describe("thirdparty", function () {
+  // Note that we're using the second account here
+  const [, wallet] = getAccounts();
+  const provider = getDefaultProvider("http://127.0.0.1:8545");
+  // const signer = wallet.connect(provider);
+  const baseSigner = wallet.connect(provider);
+  // Also demonstrates the nonce manager usage
```
carsonfarmer, line 29: Yeh, pretty cool. This is showcasing how someone might want to use a nonce-tracker with tableland. It is as easy as wrapping your signer with a NonceManager... nice!

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
...
+          type: DataTypes.STRING,
+        },
+      }
+    );
+    type User = Infer<typeof users>;
+
+    this.beforeAll(async function () {
+      this.timeout("10s");
+
+      const create = await users.CreateTable({
```
carsonfarmer, line 65: Even creating tables works?! Rad. In this case though, we need a slight "hack" to update the tableName, which is marked as immutable (though can be manipulated if we force it).

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      deepStrictEqual(results, [
+        {
+          name: "Bobby Tables",
+          id: 1,
+          email: "bobby-tab@gmail.com",
+        },
+      ]);
+    });
+
+    test("basic query building works well to then query the data", async function () {
```
carsonfarmer, line 109: This query builder pattern is kinda nice. There are others that also work with D1, but this one is the most advanced IMO.

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      const results = await db
+        .prepare(`SELECT * FROM ${users.tableName} WHERE id=?`)
+        .bind(user.id)
+        .first<User>();
+      deepStrictEqual(results, user);
+    });
+  });
+
+  describe("@databases/sql", function () {
```
carsonfarmer, line 177: This is a lib that I've seen others using. It is quite nice, and has lots of interesting functionality. In particular, using tagged template literals for working with SQL is really nice. I'd go so far as suggesting we recommend this in our docs. I'll mention why in a sec.

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    // See https://www.atdatabases.org/docs/sqlite
+    const sqliteFormat: FormatConfig = {
```
carsonfarmer, line 179: We support other types of placeholders, so we could include those here if we wanted. This is taken directly from the @databases/sqlite source.

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      formatValue: (value) => ({ placeholder: "?", value }),
+    };
+    let tableName: string;
+
+    this.beforeAll(async function () {
+      this.timeout("10s");
+
+      // First, we'll test out using sql identifiers
+      const primaryKey = sql.ident("id");
+      const query = sql`CREATE TABLE test_sql (${primaryKey} integer primary key, counter integer, info text);`;
```
carsonfarmer, line 190: One nice thing about tagged template literals like this, is that you can actually get syntax highlighting. There is a plugin for VSCode, so we could also recommend that.

test/thirdparty.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    });
+
+    test("inserting rows with interpolated values is possible", async function () {
+      const one = 1;
+      const four = 4;
+      const three = sql.value("three");
+      // Here's a safer way to inject table names
+      const query = sql`INSERT INTO ${sql.ident(
+        tableName
+      )} (counter, ${sql.ident("info")})
```
carsonfarmer, line 204: Here's where this pattern shines. We can _safely_ interpolate identifiers into a query string using this library. I'm actually half tempted to do this in our own source using this lib. This is much nicer than just string interpolation, and safer because it'll escape properly. The same goes for values, and just about anything else you interpolate into the tagged template literal. It remain to be seen if we can auto-encode things like dates and blobs... I think it'll work.

test/validator.test.ts
```diff
@@ -0,0 +1,371 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
carsonfarmer, line 1: Just the tests for the core validator API. You'll notice that I opted _not_ to include unit tests for the individual API functions. I _had_ those tests in there, but felt it was overkill given the 100% test coverage we get with this higher-level set of tests. But I could probably be convinced to test the lower level APIs as well if we wanted it.

test/validator.test.ts
```diff
@@ -0,0 +1,371 @@
...
+  getDefaultProvider,
+  getBaseUrl,
+  getChainId,
+  overrideDefaults,
+} from "../src/helpers/index.js";
+import { Database } from "../src/index.js";
+import { Validator } from "../src/validator/index.js";
+import type { WaitableTransactionReceipt } from "../src/registry/index.js";
+
+const chainId = getChainId("localhost");
```
carsonfarmer, line 22: Just for fun, and to showcase this particular helper function.

tsconfig.json
```diff
@@ -1,27 +1,15 @@
 {
```
carsonfarmer, line 1: Just the bare minimum here. You'll note that I'm targeting ES2020, with module set to ES 2022. This is to match D1 lib, which supports errors with a cause, which is an ES 2022 thing in typescript. Additionally, we are supporting DOM APIs and ES 2022 features. Note that I'm only including source here in the build. The fun thing about mocha is that it will auto-compile our tests even if they aren't included here (though we do include tests in our lint tsconfig).

carsonfarmer (commented):
Second pass at review comments for @joewagner, and perhaps @asutula or @awmuncy.

src/database.ts
awmuncy: Dumping would be pretty useful for any pre-write checks folks want to do to validate that their transaction is not only valid, but doesn't conflict with the network state of the table. (UNIQUE conflicts for example)

awmuncy (commented):
I really like the layout of the project now. Registry, validator, helpers, tests. The was a herculean effort and it looks sharp.

src/helpers/binding.ts
```diff
@@ -0,0 +1,163 @@
...
+    case "number":
+      return bindNumber(param);
+    case "boolean":
+      return bindBoolean(param);
+    case "string":
+      return bindString(param);
+    case "undefined":
+      return bindNull(param);
+    case "object":
+      if (param instanceof Date) {
```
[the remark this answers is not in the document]
carsonfarmer: Yeh I considered this. But to be honest, there didn't seem like a good reason to "abuse" that pattern here. If/else are good for boolean comparisons (which have have here because of JavaScripts strange typeof/instanceof behaviors). It is true that switches are nicer for fixed comparisons (which we have here), but if you "read" the if/else branches in plain English, it is quite clear what is happening here.

src/helpers/binding.ts
```diff
@@ -0,0 +1,163 @@
+const placeholderRegExp =
+  /["'`](?:(?<=")[^"\\]*(?:\\.[^"\\]*)*"|(?<=')[^'\\]*(?:\\.[^'\\]*)*'|[^`]*`)|(\?\d*|[:@$][a-zA-Z_]\w+)/gmu;
```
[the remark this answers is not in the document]
carsonfarmer: Yeh, so you'll note that this is using non-capturing groups for al the "escape" items (", ', `), and lookbehind to make sure that we're capturing everything between a matching set of either ", ', or `. The secondary quotes are all |'rd together _with_ the placeholder items (?, ?NN, :, @, %), so that we match _either_ an escaped quote, _or_ one of the placeholders. If we match on the string escapes, we short circuit the |s and we won't find them. This works so that we don't try to replace quoted ? and things like that. I'm sure there are edge cases that this will miss (for instance, I've now noticed I also have to add [] escaping), but give we'll eventually have real placeholder treatment and prepared statements, I think this works ok for now.

src/validator/client/index.ts
```diff
@@ -0,0 +1,14 @@
+import { Fetcher } from "./fetcher";
+import { ApiResponse, ApiError, FetchConfig } from "./types";
```
awmuncy, line 2: Need to add `.js` extension to imports here:
 
```
import { Fetcher } from "./fetcher.js";
import { ApiResponse, ApiError, FetchConfig } from "./types.js";
```

src/validator/client/fetcher.ts
```diff
@@ -0,0 +1,272 @@
...
+  FetchConfig,
+  Method,
+  Middleware,
+  OpArgType,
+  OpenapiPaths,
+  OpErrorType,
+  Request,
+  _TypedFetch,
+  TypedFetch,
+} from "./types";
```
awmuncy, line 19: Need to add `.js` extension to imports here:
 
```
import {
  ApiError,
  ApiResponse,
  CreateFetch,
  CustomRequestInit,
  Fetch,
  FetchConfig,
  Method,
  Middleware,
  OpArgType,
  OpenapiPaths,
  OpErrorType,
  Request,
  _TypedFetch,
  TypedFetch,
} from "./types.js";
```

src/validator/client/index.ts
carsonfarmer: Was this causing issues for you? If not, these come from an upstream lib and I'd like to avoid modifying the code too much if we can avoid it. If it is causing issues, then that's another thing and we'll have to do that.

awmuncy: Yes, unfortunately it was causing issues.

carsonfarmer: Bummer, ok!

src/validator/client/validator.ts
```diff
@@ -0,0 +1,288 @@
...
+       * @description The parent chain to target
+       * @example 80001
+       */
+      /**
+       * @description Table identifier
+       * @example 1
+       */
+      readonly path: {
+        chainId: number;
+        tableId: string;
```
awmuncy, line 268: Why is the `tableId` a string instead of a number?

carsonfarmer: Because large integers... It should now be a string across the entire stack, save perhaps for the smart contract itself. But the cool thing there is that ethersjs has this nice "big int ish" type that supports strings and ints and bigints. Anyway, long story short, if you want consistent big ints, use strings.

carsonfarmer:
Hey @joewagner and @awmuncy the tests and everything are all passing again. I have made several updates since this was initially PR'd, so you might want to just review it fresh at this point? Ideally we get this reviewed fully by at least one of you before the end of the week!

joewagner:
> Hey @joewagner and @awmuncy the tests and everything are all passing again. I have made several updates since this was initially PR'd, so you might want to just review it fresh at this point? Ideally we get this reviewed fully by at least one of you before the end of the week!

Sounds great!

src/statement.ts
joewagner: I've been testing this, and it probably should have been obvious at first ;) but
```
const res = await db.prepare(`SELECT * FROM ${tableName};`).first();
console.log(res);
// { col: "val1" }
```
and
```
const res = await db.prepare(`SELECT * FROM ${tableName};`).all();
console.log(res);
/*
{
  meta: { duration: 17.52889347076416 },
  success: true,
  results: [ { col: 'val1'}, { col: 'val2' } ]
}
*/
```

Basically first returns a single row as an object and all returns an array with each row as an object.  Dobule checking I understand this correctly?

src/statement.ts
joewagner: This name feels counter intuitive to me, e.g. I run a query to see results, but I'm guessing this is how the D1 api works?

src/statement.ts
joewagner: putting a note here for my own reference. example response is:
```
await db.prepare(`SELECT * FROM ${queryableName};`).raw();
// => [ [ 'tree', 'aspen' ], [ 'tree', 'pine' ] ]
```

src/validator/receipt.ts
```diff
@@ -0,0 +1,85 @@
...
+    .method("get")
+    .create();
+  const { data } = await receiptByTransactionHash(params, opts).catch(
+    hoistApiError
+  );
+  const transformed = transformResponse(data);
+  return transformed;
+}
+
+export async function pollTransactionReceipt(
```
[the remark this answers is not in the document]
joewagner: @carsonfarmer We talked about this in the voice channel, but just putting a note here that the polling seems to keep happening for 6 iterations even if the materialization is confirmed on the first request.  Not exactly a critical bug, but it would be nice skip the unneeded http requests.

src/statement.ts
carsonfarmer: Yes your interpretation is correct. It is really just a shortcut for puling out the first result. Honestly kinda seems like a silly method. In theory, if we were using real prepared statements, then this method makes sense, because you'd just retrieve the first found record.

src/statement.ts
carsonfarmer: Yeh, run is the side-effects version, comes from D1, better-sqlite3, etc

src/statement.ts
carsonfarmer: Yes, raw is essentially what our current "table" version returns, minus the extra columns field with the column names.

src/validator/receipt.ts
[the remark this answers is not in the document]
carsonfarmer: Ooof, I totally forgot about this. I just pushed an update/fix and some tests to check for this. Thanks for reminding me!

awmuncy (approved):
Been poking around and using the SDK rewrite for a while and it is a pleasure to work with. Just gave it another once over. It looks good to go to me, absolutely love that we've got this compatibility now. 

LGTM.

joewagner (approved):
Overall this is gonna be an amazing improvement.  I have a few implementation questions as inline comments.
I also made an inline comment about building for cjs if we want to allow hardhat plugins to use the new sdk. (which I think we do)

.github/workflows/combine-prs.yml
```diff
@@ -11,7 +11,7 @@ on:
       mustBeGreen:
         description: "Only combine PRs that are green (status is success)"
         required: true
-        default: true
+        default: "true"
```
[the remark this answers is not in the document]
carsonfarmer: Hmmm, not sure. I think my linter just did this? I can revert for sure.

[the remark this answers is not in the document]
carsonfarmer: Yeh, I have a GitHub actions yaml linter, and it's saying that true (no quotes) isn't the right type. So we should probs update the others if/when we can.

.github/workflows/publish.yml
```diff
@@ -20,8 +21,8 @@ jobs:
             ${{ runner.os }}-build-
             ${{ runner.os }}-
 
-      - name: Setup Node Environment ⬢
-        uses: actions/setup-node@v1
+      - name: Setup ⬢
+        uses: actions/setup-node@v2-beta
         with:
           node-version: 16
```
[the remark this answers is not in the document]
carsonfarmer: Yeh, let's update to 18 here, and then do a full matrix strategy update for all of them as separate work.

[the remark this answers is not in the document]
carsonfarmer: Oh wait, actually we only need a matrix on review and test, not publish and docs. So we can just add 18 to the matrix on test and review and bump this and docs to 18.

src/database.ts
```diff
@@ -0,0 +1,97 @@
...
+        );
+      } else {
+        // Mutating queries are sent as a single query to the smart contract
+        const sql = statements.map((stmt) => stmt.toString()).join(";");
+        const result = await this.prepare(sql).all<T>(undefined, opts);
+        return [result];
+      }
+    } catch (cause: any) {
+      if (cause.message === "ALL_ERROR") {
+        throw new Error("BATCH_ERROR", { cause: cause.cause });
```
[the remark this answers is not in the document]
carsonfarmer: Yeh, I know this is annoying. They need to extract the cause from the Error type (see the README). This is how D1 does it. You know what... I'll do a follow up PR that combines the two approaches. We'll still be mostly D1 compatible, but with more informative top level error messages.

src/database.ts
carsonfarmer: Yeh see my comment to your previous message. I think we can do exactly what you have above. In fact, truth be told, during development that's exactly how I had it because those error codes were super annoying for testing envs as you mention above!

src/helpers/binding.ts
carsonfarmer: Well, ideally no one is reading this as it isn't part of the public API. But yeh good call. I'll add a comment.

.github/workflows/combine-prs.yml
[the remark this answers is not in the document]
joewagner: Seems like that input is supposed to be a boolean, not a string. I _think_ the check for mustBeGreen on line 52 is always going to pass if this is a string.  With that said I don't know enough about `actions/github-script@v6` to say for sure.
What does your linter do if you try putting it back to `default: true` and add `type: boolean`?
cc: @asutula

src/statement.ts
```diff
@@ -0,0 +1,240 @@
...
+      const start = performance.now();
+      const { sql, type, tables } = await this.#parseAndExtract();
+      switch (type) {
+        case "read": {
+          const config = await extractReadonly(this.config, {
+            type,
+            tables,
+          });
+          await queryAll(config, sql, opts);
+          return wrapResult([], performance.now() - start);
```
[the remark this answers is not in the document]
carsonfarmer: I don't think, if they've read the docs correctly, anyone would use run with a read query. But since it is possible for them to do it, I went with this. This is the behavior that D1 follows. It'll still run the query, but it doesn't return the result. I guess it might be useful for timing the response? Not really sure.
- I don't think logging a warning is needed, since this behavior is already documented.
- I don't like the idea of adding an extra parameter here, just for this.
- But there _will_ be results in the case that the read query was valid. When we call wrap results, it'll take that and include it in the output.

Anyway, I'm ok with us just _not_ throwing away the returned response in this case (and including it as the "results" object), just because it seems slightly more intuitive if you did give a ready query to the run statement. The only downside is that this does _not_ match the D1 behavior. I guess that's ok? As it can simply be ignored by downstream consumers. I get that it seems silly to even expect a response at all in this case.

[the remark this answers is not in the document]
carsonfarmer: Thinking about this a bit more, I still don't think we want an extra flag here on the API, because we only have a read API. So there is really no point in calling it if you don't want results. I had originally throw an error when you tried to run a read query, but D1 does actually execute the query it seems, and returns an empty result. So I went with that. But we can just include the results here if that feels more intuitive. Long story short though: don't use run for reads.

src/database.ts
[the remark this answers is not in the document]
joewagner: A follow up PR sounds great!

src/statement.ts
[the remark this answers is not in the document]
joewagner: I guess leaving it as is makes the most sense, good docs will make this a non-issue.

src/database.ts
carsonfarmer: Ok, yeh its a quick change to include more context at the top-level error message. So added.

src/database.ts
carsonfarmer: What do you think of my approach here @joewagner ? Does this seem reasonable for now?

joewagner: Yeah, this looks good for now.  The long outstanding [runSQLs PR](https://github.com/tablelandnetwork/evm-tableland/pull/257) in evm-tableland to enable an array of statements will be pretty relevant to this once its finished.  I'l try to prioritize that and have it rfr by next week. 'Till then this looks good!

src/validator/client/validator.ts
```diff
@@ -0,0 +1,288 @@
+/**
```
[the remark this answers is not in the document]
carsonfarmer: What is your opinion here @joewagner?

src/validator/query.ts
```diff
@@ -0,0 +1,44 @@
...
+  columns: Array<{ name: string }>;
+}
+
+export type ObjectsFormat<T> = T[];
+
+export type BaseParams = Paths["/query"]["get"]["parameters"]["query"];
+export type Format = BaseParams["format"];
+export type Params<T extends Format> = BaseParams & { format?: T };
+
+export async function getQuery<T = unknown>(
```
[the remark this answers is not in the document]
carsonfarmer: Have you seen any nicer examples of this type of thing @joewagner?

test/statement.test.ts
carsonfarmer: I actually dislike it enough that I'm glad we didn't do it here.

src/validator/client/validator.ts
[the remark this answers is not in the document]
joewagner: What you have here works well for me. Can prettier and the linter ignore the file?

src/validator/query.ts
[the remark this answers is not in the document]
joewagner: No, this kinda thing seems tricky and since you have something working I'm all for using it.

src/helpers/connection.ts
carsonfarmer: Narrator, he did rename it.

src/validator/client/validator.ts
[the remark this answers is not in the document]
carsonfarmer: Yeh we certainly could. Yeh we probably should.

.github/workflows/combine-prs.yml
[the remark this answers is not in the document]
carsonfarmer: yup, the linter links that type: boolean @joewagner. nice catch!

src/database.ts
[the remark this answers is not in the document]
carsonfarmer: Naw I just snuck it in here.

src/statement.ts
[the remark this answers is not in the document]
carsonfarmer: Well I went and made it so that it'll return the results if it is a read. I think that's fine. No sense in wasting an API call right? If it turns out to be a compatibility issue, we can revert. But we'll make sure the docs say the right thing (which the API.md doc does).

carsonfarmer:
Ok @joewagner, I think I've addressed all comments. I've also added support for cjs builds, even though I am opposed to this . For the sake of hardhat, I'll comply 😜

joewagner:
> Ok @joewagner, I think I've addressed all comments. I've also added support for cjs builds, even though I am opposed to this . For the sake of hardhat, I'll comply 😜

haha, hardhat is thankful 🤣 
This all looks good enough to land on main!

PR #331, opened 2022-12-21
Lint and pretty fixes + tests use next validator

joewagner:
@carsonfarmer I got the tests passing with a build of the validator that includes the testnet/mainnet split.  I also fixed the lint and prettier errors

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,1 @@
+coverage/*
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,0 @@
-coverage/*
--- a/src/validator/client/README.md
+++ b/src/validator/client/README.md
@@ -2,7 +2,7 @@
 
 # 📘️ openapi-typescript-fetch
 
-A typed fetch client for [openapi-typescript](https://github.com/drwpow/openapi-typescript) 
+A typed fetch client for [openapi-typescript](https://github.com/drwpow/openapi-typescript)
 
 **Features**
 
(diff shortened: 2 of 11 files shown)


.eslintignore
```diff
@@ -0,0 +1 @@
+coverage/*
```
carsonfarmer, line 1: Rather that having another admin file, I was opting for "ignorePatterns" in the eslintrc.json file. I'm going to merge this in, and then revert to using that if you don't mind?

src/validator/client/README.md
```diff
@@ -2,7 +2,7 @@
 
 # 📘️ openapi-typescript-fetch
 
-A typed fetch client for [openapi-typescript](https://github.com/drwpow/openapi-typescript) 
+A typed fetch client for [openapi-typescript](https://github.com/drwpow/openapi-typescript)
```
carsonfarmer, line 5: This whole package is actually "vendor" from a separate project, so I had opted to _not_ lint or format it. Actually, it should have been added to the ignored patterns for prettier and eslint. Buuuut, it looks like all the issues are auto-fixable, so I'll allow it in this case 😉 . Good idea.

carsonfarmer (approved):

carsonfarmer:
I hope you don't mind @joewagner, but I'm going to merge this PR into my branch. Thanks for the fixes!

joewagner:
> I hope you don't mind @joewagner, but I'm going to merge this PR into my branch. Thanks for the fixes!

@carsonfarmer perfecto.
Nice work on the refactor it's looking good!!

PR #339, opened 2023-01-02
chore: test against latest validator

carsonfarmer:
This is a "legacy snapshot" of our current SDK. It is being tested against the latest validator APIs, and all tests are passing. We can keep this branch alive for a period of time after the switch that corresponds with the period of time we keep the existing JSON RPC APIs alive on the validator. This will provide some time for folks to make the SDK switch. We can continue to publish patch updates on this branch to the 3.x version, while core fixes and new features/updates happen on the 4.x version (which will become the new main branch).

cc @joewagner (who's good idea this was)

I'm leaving this in draft mode because we don't actually want to merge this into main.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@babel/preset-env": "^7.19.3",
         "@babel/preset-typescript": "^7.16.7",
         "@babel/register": "^7.18.9",
-        "@tableland/local": "^0.0.4",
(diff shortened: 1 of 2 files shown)


joewagner:
Looks great!

> cc @joewagner (who's good idea this was)

Definitely a team effort!

joewagner:
@carsonfarmer no hurry, but I think we can close this now.

PR #352, opened 2023-01-11
Update dev deps post refactor

carsonfarmer:


joewagner (approved):
Looks good!

joewagner:
Actually I may have spoken too soon, looks like the regex assertions aren't passing.
Did you mean for them to be `/Th\w+ operation was aborted/`?

carsonfarmer:
> Actually I may have spoken too soon, looks like the regex assertions aren't passing.
> Did you mean for them to be `/Th\w+ operation was aborted/`?

See my follow up commit :P

PR #356, opened 2023-01-13
Final dep updates for pre-release cycle

carsonfarmer:


test/validator.test.ts
```diff
@@ -330,6 +330,83 @@ describe("validator", function () {
       deepStrictEqual(response.rows, [[1]]);
     });
 
+    test("where query is called with unwrap as 'true' on multiple rows", async function () {
```
carsonfarmer, line 333: I snuck a few extract tests in here, as I pulled them _out_ of @tableland/evm, and refactored them into here. Just because we don't really expose these options directly in the SDK, and there was no point in really testing them over there. But here, it is useful to test the failure cases that I wasn't testing previously.

awmuncy (approved):

PR #364, opened 2023-01-18
Trim whitespace from create statements

awmuncy:
Fixes bug where whitespace at the beginning of a create statement prevents the SDK from adding the chainID to the create statement before sending it to the registry. 

## Checklist:

### DIFF
--- a/src/registry/create.ts
+++ b/src/registry/create.ts
@@ -25,7 +25,6 @@ export async function prepareCreateTable({
     `${first}_${chainId}`,
     true
   );
-  statement = statement.trim();
   const stmt = statement.replace(
     firstSearch,
     function (_, create: string, name: string, schema: string) {
--- a/src/registry/create.ts
+++ b/src/registry/create.ts
@@ -25,9 +25,8 @@ export async function prepareCreateTable({
     `${first}_${chainId}`,
     true
   );
-  const stmt = statement
-    .trim()
-    .replace(
+  statement = statement.trim();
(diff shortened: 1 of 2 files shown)


src/registry/create.ts
```diff
@@ -25,6 +25,7 @@ export async function prepareCreateTable({
     `${first}_${chainId}`,
     true
   );
+  statement = statement.trim();
```
carsonfarmer, line 28:
```suggestion
  // Trim whitespace to avoid issues with regex later
  statement = statement.trim();
```

carsonfarmer (approved):

carsonfarmer (changes requested):
Oh second thought, let's just fix the regex right away, or better still, based on a convo with @joewagner he suggested we should just trim all statements within the Statement constructor.

joewagner:
> Oh second thought, let's just fix the regex right away, or better still, based on a convo with @joewagner he suggested we should just trim all statements within the Statement constructor.

Double check me here, but I think if we add the trim here https://github.com/tablelandnetwork/js-tableland/blob/main/src/statement.ts#L43
It would get the regex to always work and it would have the added benefit of removing unneeded spaces from the string we send to the chain.

src/registry/create.ts
awmuncy: Removed this trim in favor of a trim in the Statement constructor.

carsonfarmer (approved):

PR #367, opened 2023-01-20
Some fixes and changes to address issues with vite (and probably other build systems)

carsonfarmer:
I'll leave this one a draft for now, as there will likely be other issues.

### DIFF
--- a/API.md
+++ b/API.md
@@ -362,6 +362,68 @@ console.log(rows[1].results);
 */
 ```
 
+## Additional APIs
+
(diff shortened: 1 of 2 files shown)


carsonfarmer:
Assuming those v16 tests pass this time around @joewagner and @datadanne , can you take a look and give me a 👍 if it seems ok? We can add more PRs before we push out another .pre update from this stuff. I've tested this with a vite react app, and it works nicely.

joewagner (approved):
@carsonfarmer this looks great to me!

PR #368, opened 2023-01-20
help get chainId in multiple environments

### LINKED ISSUE #366: Table name uses chain ID of `0` for all chains
**Describe the bug**
I'm creating and writing to a table using the SDK v4.0.0-pre.4. Upon preparing & running the create statement, I save the table name from the statement response (`meta.txn.name`) to a variable. The returned table name uses a chain ID of `0` instead of the correct / actual chain ID. 

Similarly, I insert a value into the table. To get around the issue above, the table's name is hardcoded. Upon logging the response from the executed statement, the table's name oddly has a chain ID of `0`, even though the statement passed a value of `80001` and successfully mutated the table.

This was tested on local tableland as well as polygon mumbai.

**To Reproduce**
Create a ts react app and use the `@tableland/sdk@4.0.0-pre.4`:
```
npx create-react-app test --template typescript && npm i
npm i @tableland/sdk@4.0.0-pre.4
```

Here is an example of the table create issue (e.g., use somewhere in `App.tsx`, like on a button click):
```ts
import { Database } from "@tableland/sdk";

interface Schema {
  id: number;
  name: string;
}

const db = new Database<Schema>();

const prefix: string = "my_sdk_table";
const { meta: create } = await db
  .prepare(`CREATE TABLE ${prefix} (id integer primary key, name text);`)
  .run();

let name;
if (create.txn !== undefined) {
  console.log(create.txn.name);
  name = create.txn.name;
} 
```
This will successfully create the table and assign the `create.txn.name` value to the `name` variable. However, the value of `name` is, e.g., `my_sdk_table_0_5324` instead of `my_sdk_table_80001_5324`. This is true for local tableland as well.

Then, as expected, subsequent references to this table name throw an error, such as table reads or writes:
```
Uncaught (in promise) Error: RUN_ERROR: chain id mismatch: received 0, expected 31337
```

To fully test out the write query, I hardcode the table name. Out of curiosity, I log the response from the write query, which incorrectly shows the table name with a chain ID of `0`, even though I successfully write to the table with a chain ID of `80001`:
```ts
const name = "my_sdk_table_80001_5324";
const { meta: insert } = await db
  .prepare(`INSERT INTO ${name} VALUES(?,?);`)
  .bind(1, "Tableland rocks: https://tableland.xyz/")
  .run();
await insert.wait;
console.log(insert);
```
This is the `insert` object logged:
```js
{
    duration: "18395.10000014305",
    txn: {
      blockNumber: 31166186,
      chainId: 0,
      name: "my_sdk_table_0_5324",
      prefix: "my_sdk_table",
      tableId: "5324",
      transactionHash:
        "0xea9afce1c0473bdb3e4b722b5e537208bfb80edfcc69c6a352a82f8629a75c9f",
    },
}
```

**Expected behavior**
The value of the table's `name` should have the correct chain ID, not `0`.

**Desktop (please complete the following information):**

- OS: Mac M1
- Browser: Chrome
- SDK v4.0.0-pre.4
- Chains tested: local tableland, polygon mumbai

joewagner:
fix #366 

When connecting with Metamask the chainId is not available in the same way as when connecting in a node.js environment.
This contains some alternative checks for chainId, and includes some error messaging to inform devs of connection problems.

### DIFF
--- a/src/helpers/chains.ts
+++ b/src/helpers/chains.ts
@@ -57,10 +57,19 @@ const supportedChainsById = Object.fromEntries(
  * @returns An object containing the default chainId, contractAddress, chainName, and baseUrl for the given chain.
  */
 export function getChainInfo(chainNameOrId: ChainName | number): ChainInfo {
-  if (typeof chainNameOrId === "number") {
-    return supportedChainsById[chainNameOrId];
+  const chainInfo =
+    typeof chainNameOrId === "number"
+      ? supportedChainsById[chainNameOrId]
+      : supportedChains[chainNameOrId];
+
+  // Something about the way we are using `Object.fromEntries` to generate these
+  // maps makes eslint think that this value is always true, hence the disable
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+  if (!chainInfo) {
+    throw new Error(`cannot use unsupported chain: ${chainNameOrId}`);
   }
-  return supportedChains[chainNameOrId];
+
+  return chainInfo;
 }
 
 export function isTestnet(chainNameOrId: ChainName | number): boolean {
--- a/src/registry/utils.ts
+++ b/src/registry/utils.ts
@@ -7,6 +7,7 @@ import {
   type Config,
   type ReadConfig,
   extractBaseUrl,
+  extractChainId,
 } from "../helpers/config.js";
 import {
   type ContractTransaction,
@@ -79,12 +80,16 @@ export async function extractReadonly(
 }
 
 export async function wrapTransaction(
-  conn: ReadConfig,
+  conn: Config,
   prefix: string,
   tx: ContractTransaction
 ): Promise<WaitableTransactionReceipt> {
   const params = await getContractReceipt(tx);
-  const name = `${prefix}_${params.chainId}_${params.tableId}`;
+  const chainId =
+    params.chainId === 0 || params.chainId == null
+      ? await extractChainId(conn)
+      : params.chainId;
+  const name = `${prefix}_${chainId}_${params.tableId}`;
   const wait = async (
     opts: SignalAndInterval = {}
   ): Promise<TransactionReceipt & Named> => {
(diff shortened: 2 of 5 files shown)


src/helpers/chains.ts
```diff
@@ -57,10 +57,19 @@ const supportedChainsById = Object.fromEntries(
...
+  const chainInfo =
+    typeof chainNameOrId === "number"
+      ? supportedChainsById[chainNameOrId]
+      : supportedChains[chainNameOrId];
+
+  // Something about the way we are using `Object.fromEntries` to generate these
+  // maps makes eslint think that this value is always true, hence the disable
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+  if (!chainInfo) {
+    throw new Error(`cannot use unsupported chain: ${chainNameOrId}`);
```
joewagner, line 69: This isn't really related to #366, but I accidentally configured Metamask with a chainId we don't support and the resulting error was very cryptic.

src/registry/utils.ts
```diff
@@ -79,12 +80,16 @@ export async function extractReadonly(
 }
 
 export async function wrapTransaction(
-  conn: ReadConfig,
+  conn: Config,
```
carsonfarmer, line 83: Is this going to cause issues at all? Before we were only requiring a read config. I guess this is ok, because we likely won't be wrapping a transaction at all if we are just in readOnly mode, right?

src/helpers/chains.ts
carsonfarmer: huh, ok, nice to have a better error message

carsonfarmer (approved):
Seems like good fixes? Not sure what the implication of the config requirement might do. Seems ok though.

src/registry/utils.ts
joewagner: I _think_ it should be ok.  All the tests pass and in the little demo app @dtbuchholz setup it seemed to work.

PR #384, opened 2023-02-03
Simplify regexp to avoid look-behind. fixes #383

### LINKED ISSUE #383: RegExp used in prepareCreateTable isn't supported in safari (because of lookbehind)
The RegExp used here isn't supported by Safari (🤦) which means that web apps that use the SDK don't work in Safari at all.
 
https://github.com/tablelandnetwork/js-tableland/blob/9f40679f2a7f491ccf516727d25470940797b040/src/registry/create.ts#L9-L11

https://caniuse.com/js-regexp-lookbehind

carsonfarmer:
Fixes #383. Also attempt to strip extra whitespace and keep the create statements as short as possible. If an input name includes escape chars or things like that, it will now just escape the whole thing to avoid strange half-names etc.

This passes all existing tests, and also improves the behavior of strange inputs like this:

```sql
CREATE           TABLE  something_80001           (id INT);
```

If someone tries to create a table with a name like this: `something'"[]blur ` it'll essentially convert this to `[something'"[]blur]`, which might not be expected... but is probably better than the alternative? In this particular case, this should _still_ fail, which is a good thing.

### DIFF
--- a/src/registry/create.ts
+++ b/src/registry/create.ts
@@ -6,9 +6,7 @@ import { getContractAndOverrides } from "./contract.js";
 // Match _anything_ between create table and schema portion of create statement
 const firstSearch =
   /(?<create>^CREATE\s+TABLE\s+)(?<name>\S+)(?<schema>\s*\(.*\)[;]?$)/i;
-// Match _anything_ between pairs of "", ``, '', or []
-const secondSearch =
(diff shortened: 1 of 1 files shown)


joewagner (approved):
It's pretty tough to reason about these kind of `RegExp`s, and since we only test in the node.js env, it's tough to know if this will work without setting up an application that uses the browser.  I'll approve so that we can get things working in Safari. 
@carsonfarmer Should build out a way to make testing across different browsers easier?

carsonfarmer:
Yeh that was kinda my goal here. If it is hard to know what different browsers are going to do, maybe just keep it simple and handle the issues ourselves. Anyway, you are right on that testing front. We have to get a good browser flow going. Something like Puppeteer (or whatever newer better solution is out there) would be good.

PR #394, opened 2023-02-10
Add Playwright tests that run on the browser

joewagner:
## Summary

This PR adds a very basic app that uses the SDK and is then used in Playwright tests that run on chromium(chome, edge, and others), firefox, and webkit(safari).
You can also just run the app and do manual tests with whatever browser you prefer.  
The tests are also part of CI, but I gave them their own command.  I thought about adding the command to `npm test` but decided to separate it, similar to how we separated `npm run coverage`. I'm not 100% sure on this choice, it's reversible, if anyone has different idea how to incorporate it I'm very open to alternatives.

## Details

Regarding the bug in Safari described in #393.  (Un)fortunately the webkit version is the newest release and as such it has support for the RegExp that was making the SDK unusable on Safari.  A bit of a double edged sword here. We are testing the SDK against the soon to be browser implementations, so we will be able to know ahead of time if an incoming change to the browser will cause a problem with our SDK.  However, this also means we are not testing the actual current browser releases. If someone wants to test the current Safari they can start the test app as descibed below, open a debugger, and interact with the test app manually.
If this isn't sufficient, we can potentially setup Playwright to use older versions of the browsers we want to support.

To start the app for manual use:
- Start a Local Tableland instance however you prefer.
- do `npm run test-server`.  The app will be served at http:localhost:5173

The App is a few simple forms which expose the ability to do some of the more common SDK interactions.
It is not user friendly, and really designed for use by the Playwright tests.  However, as mentioned above, it's potentially convenient to run it and test alternative browsers manually, or if you are having trouble writing working tests.
In the future it can also serve as a very quick way to try reproducing a bug report that involves a specific browser.

The App does **not** use browser wallets like metamask, and instead just uses ethers.js with a public hardhat wallet.

I only added "happy tests". We can definitely add more as needed, but for the initial purposes I am thinking we can keep the tests quick and acting more as a smoke screen than thorough protections.


joewagner:
Had to rerun the tests for some reason, but this should all be rfr

test/browser/server/src/App.vue
```diff
@@ -0,0 +1,109 @@
...
+
+        if (!statementStr) throw new Error("no statement available");
+        if (!successMsg) throw new Error("no success message available");
+
+        const res = await db
+          .prepare(statementStr)
+          .all();
+
+        this.$data.message = successMsg;
+        this.$data.responseJson = JSON.stringify(res);
```
joewagner, line 50: The tests interact with the app via a headless browser and then expect the html content to be changed.  By using a simple text message and a stringified object we can test everything we need to.

test/browser/test/happy.spec.js
```diff
@@ -0,0 +1,132 @@
...
+
+  test("can create", async function ({ page }) {
+    await page.goto("http://localhost:5173/");
+
+    const success = "table was created";
+    await page.type("input[name=statement]", "CREATE TABLE browser_table (k text, v text, num integer);");
+    await page.type("input[name=success]", success);
+    await page.click("input[name=submit-sql]");
+
+    await expect(page.getByTestId("status")).toHaveText(success);
```
joewagner, line 21: The Playwright `expect` test assertion works internally by constantly polling the page's html until either it matches the expectation, or the config timeout is reached.

test/browser/test/happy.spec.js
```diff
@@ -0,0 +1,132 @@
...
+
+    const success = "table was read";
+    await page.type("input[name=statement]", `SELECT * FROM ${tableName};`);
+    await page.type("input[name=success]", success);
+    await page.click("input[name=submit-sql]");
+
+    await expect(page.getByTestId("status")).toHaveText(success);
+
+    const responseStr = await page.getByTestId("response").textContent();
+    const responseObj = JSON.parse(responseStr);
```
joewagner, line 64: The app writes the stringified js object to the dom and then the test can parse it an make sure it's what we expect.

test/browser/test/setup.js
```diff
@@ -0,0 +1,19 @@
+// NOTES:
+//  - Playwright will start the app in `../server` before running tests
+//  - Playwright uses a custom loader. https://nodejs.org/docs/latest-v18.x/api/esm.html#loaders
+//    This does not work with Hardhat and Local Tableland. This may be related to https://github.com/microsoft/playwright/issues/16185
+//    As a workaround the playwright tests are run with the PW_TS_ESM_ON env var set.
+//    Which means tests can not use typescript.
+import { LocalTableland } from "../server/node_modules/@tableland/local/dist/esm/main.js";
```
joewagner, line 7: Aside from what is written in the notes comment this is pretty standard usage.

test/database.test.ts
```diff
@@ -7,6 +7,7 @@ import { Database } from "../src/database.js";
 import { Statement } from "../src/statement.js";
 
 describe("database", function () {
+  this.timeout("10s");
```
joewagner, line 10: Not related to this PR, but the mocha tests are occasionally timing out.  I moved from 5s to 10s in several of the files that were timing out.

joewagner (commented):
This is ready for review!

.github/workflows/test.yml
```diff
@@ -46,3 +50,9 @@ jobs:
 
       - name: Test/Coverage 🧪
         run: npm run coverage
+
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps
+
+      - name: Test/Browser 🧪
+        run: npm run test:browser
```
carsonfarmer, line 58: I think this is the right call, let's keep it separate

playwright.config.js
```diff
@@ -0,0 +1,92 @@
...
+      name: 'firefox',
+      use: { ...devices['Desktop Firefox'] },
+    },
+
+    {
+      name: 'webkit',
+      use: { ...devices['Desktop Safari'] },
+    },
+
+    /* Test against mobile viewports. */
```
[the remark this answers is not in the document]
carsonfarmer: I think this subset will cover us ok for now. We really just need this to make sure we aren't breaking normal browser things. I don't think we're targeting any hyper-specific browser APIs that we need to be super careful about.

carsonfarmer (approved):
Wow, this is awesome @joewagner! I love this. I made a few minor comments and nits. The only one worth worrying about is that postinstall script. Can you confirm that this is for sure needed? I prefer to avoid that because some installers won't even run those anyway.

package.json
```diff
@@ -53,13 +53,17 @@
...
   "scripts": {
+    "postinstall": "cd test/browser/server && npm install",
     "lint": "eslint '**/*.{js,ts}'",
     "lint:fix": "npm run lint -- --fix",
     "prettier": "prettier '**/*.{ts,json,sol,md}' --check",
     "prettier:fix": "npm run prettier -- --write",
     "format": "npm run prettier:fix && npm run lint:fix",
     "prepublishOnly": "npm run build",
     "test": "mocha",
+    "test:browser": "PW_TS_ESM_ON=true playwright test",
```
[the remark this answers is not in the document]
joewagner: That's a good call!  I'll remove the postinstall and add a note to the README

joewagner:
@carsonfarmer I removed the post install and added the install steps to the CI workflow. Do you want to double check me?

carsonfarmer (approved):
Lovely

PR #402, opened 2023-02-16
Create warning about vite and sqlparser

### LINKED ISSUE #391: SDK does not work with default vite build setup (needs to be added as an optimizeDeps exclusion)
The v4 SDK does not work with the default vite configuration, it messes with how we load `main.wasm` in the parser. 

It does work if the SDK is excluded from vite's [dependency pre-bundling](https://vitejs.dev/guide/dep-pre-bundling.html) feature, as we do here:
https://github.com/tablelandnetwork/rigs/blob/c0f8534a790fdec2ecc95471267094a360e7f558/animation_url/vite.config.ts#L17-L21

I think we should either figure out a workaround so that it does work with vite out of the box _or_ document this very clearly in the Usage or Install-section in the README for the SDK.

carsonfarmer:
Temp documentation of Vite build issue around sqlparser.

Partially fixes #391

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ console.log(results);
 
 ## Build Tools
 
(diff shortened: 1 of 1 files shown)


joewagner (approved):
Looks good to me!

PR #411, opened 2023-02-24
Pass more meaningful error messages.

joewagner:
There's a few places in the SDK where an incorrect call to a function, or constructor, results in a `cannot read properties of undefined` or `xyz is not a function` error.  This PR will pass a more human friendly message so devs can debug more easily.

### DIFF
--- a/src/helpers/chains.ts
+++ b/src/helpers/chains.ts
@@ -112,13 +112,16 @@ export function overrideDefaults(
   chainNameOrId: ChainName | number,
   values: Record<keyof ChainInfo, number | string>
 ): void {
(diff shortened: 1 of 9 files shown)


carsonfarmer (approved):
Nice quick one. Cheers!

joewagner:
> Nice quick one. Cheers!

@carson Thanks!!
I found a few more places that could end up throwing cryptic errors, and I added some tests to get our coverage back up.
When you get a chance I think I will need you to re-review.

carsonfarmer (approved):
Love the added tests!

PR #432, opened 2023-03-22
Use the upcoming Registry contract changes with Runnables to enable better statement batching

### LINKED ISSUE #138: [JST-14] Auto-batch write queries
**Is your feature request related to a problem? Please describe.**
I would like to use `write` and simply insert _all_ of the values that I have (e.g., [this](https://github.com/duneanalytics/spellbook/blob/main/models/addresses/cex_addresses.sql) or something larger), but Tableland imposes a 35kb query limit. This means that I have to write my own methods for reducing some set of query values into 35kb groups, and then calling `write` multiple times with each of these groups.

**Describe the solution you'd like**
1. Create a new method that takes the table's name and the value tuples. Under the hood, this will enable the query chunking and also perform the set of writes:
`writeBatch('table_1_1', [(1, 'value_01') , (2, 'value_02')])`
2. Alternatively, maybe `write` could perform this instead. Assumes multiple statements are delimited by a `;` -- chunk these statements into 35kb queries and perform writes one after another. Or, assume there is just one statement like `INSERT INTO table VALUES (1, 'value_01') , (2, 'value_02')` -- break the VALUES tuples up into ~35kb groups, then perform the writes:
`write(statement, { mode: 'batchInserts' })` or `write(statement, { mode: 'batchValues' })`

**Describe alternatives you've considered**
Implementing the batching logic on my own.

**Additional context**
This would make it easier to write queries and think less about limits at the client level; lower onboarding friction.

<sub>[JST-14](https://linear.app/tableland/issue/JST-14/auto-batch-write-queries)</sub>

joewagner:
## Overview
This is a somewhat large PR, and it also has some far reaching implications.  Definitely carve some time out to look through it thoroughly.
The core of the changes are an update to the SDK to take advantage of the [new Registry Contract methods](https://github.com/tablelandnetwork/evm-tableland/pull/257) that enable sending a list of sql statements that will get executed in an all-or-nothing manner inside a single transaction.  There are also some additions to the code comments and small refactors.

**While reviewing keep in mind that our goal here is to only do a minor version bump.**
I have attempted to meet that criteria, but with such a large change there's a chance I missed something.

## Details
The best place to start looking at these changes is in the `database.ts` `batch()` function and the associated tests.
One of the key concepts needed to understand the intended usage of `batch` is that all of the Statements (capital S) that are being batched must be considered to have the same "type" when normalized by the sql-parser and they may only affect a single table.  This limitation serves a few different purposes.  The first is that the complexity of parsing user input is significantly reduced, which will make it much easier to reduce corner case bugs. The second is that the users of the SDK are given a more direct and simple mapping of the SDK API to the underlying mechanics that they are triggering.  Some examples are appropriate to explain further.

correct usage examples:
```
await db.batch([
    db.prepare(`create table my_table_1 (val int);`),
    db.prepare(`create table my_table_2 (name text);`),
    db.prepare(`create table my_table_3 (id text);`)
]);

await db.batch([
    db.prepare(`insert into my_table_1_11155111_123 (val) values (1);`),
    db.prepare(`insert into my_table_2_11155111_124 (name) values ('foo');`)
]);

await db.batch([
    db.prepare(`grant insert on my_table_1_11155111_123 to '0xadf000123';`),
    db.prepare(`grant insert on my_table_2_11155111_124 to '0xadf000123';`)
]);

await db.batch([
    db.prepare(`select * from my_table_1_11155111_123;`),
    db.prepare(`select * from my_table_2_11155111_124;`)
]);

```
incorrect usage examples:
```
await db.batch([
    db.prepare(`create my_table_4 (a int);`),
    // note: I don't know the tableId so I'm guessing the name
    db.prepare(`insert into my_table_4_11155111_126 (a) values (1);`)
]);

const res = await db.batch([
    db.prepare(
        `insert into my_table_1_11155111_123 (val) values (1);insert into my_table_2_11155111_124 (name) values ('foo');`
    ),
    db.prepare(`insert into my_table_3_11155111_125 (id) values ('100');`)
]);

const res = await db.batch([
    db.prepare(`insert into my_table_1_11155111_123 (val) values (1);`),
    // Am I getting results before the tx or after?
    db.prepare(`select * from my_table_1_11155111_123;`)
]);
```

The internal logic has the opportunity to break these statements apart and recombine them based on the tableId and based on statement type. After exploring how this would work in practice it became overly complex. Also, it seemed like the only real advantage was enabling SDK users to combine statements in ways that might have confusing responses and generally be better done by initially combining based on tableId in the Statement.

Of note is that the batching of reads has been unchanged in this PR, if that is something we want to optimize, I'll suggest that we do that work in a separate PR.

fixes #138

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -1,6 +1,6 @@
 import { type NormalizedStatement } from "@tableland/sqlparser";
 import { type Result, type Runnable } from "./registry/index.js";
-// import { wrapResult, wrapBatch } from "./registry/utils.js";
+import { wrapResult } from "./registry/utils.js";
 import {
   type Config,
   type AutoWaitConfig,
@@ -75,9 +75,9 @@ export class Database<D = unknown> {
   /**
    * Execute a set of Statements in batch mode.
    * Batching sends multiple SQL statements inside a single call to the
-   * network. This can have a huge performance impact, as it reduces the
-   * total number of transactions sent to the Tableland smart contract,
-   * thereby reducing gas costs.
+   * network. This can have a huge performance impact, as it only sends
+   * one transaction to the Tableland smart contract, thereby reducing
+   * gas costs.
    * Batched statements are similar to SQL transactions. If a statement
    * in the sequence fails, then an error is returned for that specific
    * statement, and it aborts or rolls back the entire sequence.
@@ -87,6 +87,7 @@ export class Database<D = unknown> {
    */
   async batch<T = D>(statements: Statement[], opts: Signal = {}): Promise<any> {
     try {
+      const start = performance.now();
       // each statement in `statements` could potentially be a ; separated set of statements
       // we need to split them up and then recombine them first as "read" and "mutate", then
       // for the mutates, combine based on the table they are mutating.  This sets us up to
@@ -100,49 +101,45 @@ export class Database<D = unknown> {
         .reduce((a, b): any => (a === b ? a : null));
       if (type == null) {
         throw new Error(
+          // TODO: should this say "e.g., create, write, read, acl"?
           "statement error: batch must contain uniform types (e.g., CREATE, INSERT, SELECT, etc)"
         );
       }
 
-      // because of the above check if the first type is the same as all the others
-      if (normalized[0].type === "read") {
+      if (type === "read") {
         return await Promise.all(
           statements.map(async (stmt) => await stmt.all<T>(undefined, opts))
         );
       }
 
-      if (normalized[0].type === "create") {
+      if (type === "create") {
         let receipt = await execCreateMany(
           this.config,
           statements.map((stmt) => stmt.toString())
         );
+
         if (this.config.autoWait ?? false) {
+          // wait until validator has materialized tables
           const waited = await receipt.wait();
           receipt = { ...receipt, ...waited };
         }
 
-        // TODO: do we need to map the tableIds back to the original statements?
-        return receipt;
+        // TODO: we need to map the tableIds back to the original statements
+        return wrapResult(receipt, performance.now() - start);
       }
 
-      if (normalized[0].type === "acl") {
+      if (type === "acl") {
         // TODO: handle batched acl statements...
         throw new Error("batch ACL statements not implemented");
       }
 
-      // if type is not read or create, then we use mutate.
-      // For mutating queries each table's statements are put into the same Runnable.
-      // To do this we will flatten all of the arrays of normalized statements, then group by table name.
+      if (type !== "write") {
+        // this should never be thrown, but check in case of something unexpected
+        throw new Error("invalid statement type");
+      }
 
-      // TOOD: recombining like this might be a bad idea, because of the single statement size limit
-      //    If we let the runnables be 1 to 1 with the caller's statements it gives the caller control
-      //    over how things are done.
-      //    BUT.... someone might try to do something like `insert into tbl1 (1); insert into tbl2 (2);`
-      //    as one of the statement strings, so we do need to do some spliting on a per statement string
-      //    basis.
-      // We can go through each of the normalized entries and see if there is more than one `table`
-      // if so split the statements array into a statement+tableId for each
-
+      // NOTE: there is a requirement that each statement is only affecting one table.
+      // If a caller wants to affect 2 tables, they can batch 2 statements.
       const runnables = (
         await Promise.all(
           normalized.map(async function (norm) {
@@ -151,15 +148,16 @@ export class Database<D = unknown> {
         )
       ).flat();
 
-console.log(runnables);
-
       let receipt = await execMutateMany(this.config, runnables);
       if (this.config.autoWait ?? false) {
         const waited = await receipt.wait();
         receipt = { ...receipt, ...waited };
       }
 
-      return receipt;
+      // TODO: Note that this method appears to be the wrong return type for D1-ORM compatability.
+      //    This includes the ability to wait for the transaction to finish, and things like the
+      //    table name, which are not what D1-ORM expects.
+      return wrapResult(receipt, performance.now() - start);
     } catch (cause: any) {
       if (cause.message.startsWith("ALL_ERROR") === true) {
         throw errorWithCause("BATCH_ERROR", cause.cause);
@@ -220,19 +218,13 @@ async function normalizedToRunnables(
 ): Promise<Runnable[]> {
   if (normalized.type !== "write") {
     throw new Error(
-      "converting to runnable is only possible for write statements"
+      "converting to runnable is only possible for mutate statements"
     );
   }
-  if (normalized.statements.length > 1) {
-    // if there's more than one statement we need a separate runnable for each
-    return (
-      await Promise.all(
-        normalized.statements.map(async function (statement) {
-          const norm = await normalize(statement);
-          return await normalizedToRunnables(norm);
-        })
-      )
-    ).flat();
+  if (normalized.tables.length > 1) {
+    throw new Error(
+      "each statement can only touch one table. try batching statements based on the table they mutate."
+    );
   }
 
   const { tableId } = await validateTableName(normalized.tables[0]);
@@ -240,7 +232,9 @@ async function normalizedToRunnables(
   return [
     {
       tableId,
-      statement: normalized.statements[0],
+      statement: normalized.statements.join(";"),
+      // TODO: the type called "write" is indicative of what we are calling "mutate" elsewhere
+      //       it would require a change to the sqlpraser, but maybe we should replace write with mutate?
       type: "write",
     },
   ];
--- a/src/database.ts
+++ b/src/database.ts
@@ -90,13 +90,11 @@ export class Database<D = unknown> {
   //    enables compatability with packages built to exend D1, then the return type
   //    here will potentially affect if/how those packages work.
   //    D1-ORM is a good example: https://github.com/Interactions-as-a-Service/d1-orm/
-  //    See this package's `thirdparty` tests for examples.
-  //    We also have to balance that with the ability of SDK user's to do things like
-  //    wait for the transaction to finish, and get the table name, prefix, etc...
   async batch<T = D>(
     statements: Statement[],
     opts: Signal = {}
-    // reads returns an Array, everything else a single result
+    // reads returns an Array with legnth equal to the number of batched statements,
+    // everything else a single result wrapped in an Array for backward compatability.
     // TODO: In order to work around the Validator API not returning all of the tableIds
     //       and continuing to work in a backward compatable way, it seems that we have to
     //       make this type `any` :(
--- a/src/database.ts
+++ b/src/database.ts
@@ -4,6 +4,7 @@ import { wrapResult } from "./registry/utils.js";
 import {
   type Config,
   type AutoWaitConfig,
+  checkWait,
   extractBaseUrl,
   type ChainName,
   getBaseUrl,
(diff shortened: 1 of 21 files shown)


src/database.ts
```diff
@@ -72,42 +76,93 @@ export class Database<D = unknown> {
...
         statements.map(async (stmt) => await normalize(stmt.toString()))
       );
+
       const type: string | null = normalized
         .map((stmt) => stmt.type)
         .reduce((a, b): any => (a === b ? a : null));
       if (type == null) {
         throw new Error(
-          "statement error: batch must contain uniform types (e.g., CREATE, INSERT, SELECT, etc)"
+          "statement error: batch must contain uniform types (i.e. one of: create, write, read, acl)"
```
joewagner, line 117: Not sure if this makes sense, but since the parser has a specific set of types I thought maybe it made sense to use that nomenclature.
I'm open to keeping the error message the same if anyone has a preference.

src/database.ts
```diff
@@ -72,42 +76,93 @@ export class Database<D = unknown> {
...
         );
       }
+
+      // "read" statement types are the simple case, we just do each of the queries
+      // and return an Array of the query results.
```
joewagner, line 122: If we want to do some kind of more efficient read batching I think we should do that in a separate PR, just so this PR doesn't get any bigger.

src/database.ts
```diff
@@ -72,42 +76,93 @@ export class Database<D = unknown> {
...
+        const receipt = await checkWait(
+          this.config,
+          await execCreateMany(
+            this.config,
+            statements.map((stmt) => stmt.toString())
+          )
+        );
+
+        // TODO: wrapping in an Array is required for back compat, consider changing this for next major
+        return [wrapResult(receipt, performance.now() - start)];
```
joewagner, line 140: Since this is a single transaction, we _could_ return an Object instead of an Array, but I'm not sure if that breaks D1 compatibility.

src/database.ts
```diff
@@ -162,3 +217,33 @@ export class Database<D = unknown> {
...
+ * used in a call to the registry contract.
+ * @param normalized A normalized statement, e.g. what is returned from the parser's normalize function
+ * @returns An Array of Runnables
+ */
+async function normalizedToRunnables(
+  normalized: NormalizedStatement
+): Promise<Runnable[]> {
+  if (normalized.type !== "write" && normalized.type !== "acl") {
+    throw new Error(
+      "converting to runnable is only possible for mutate statements"
```
joewagner, line 232: note: the term "mutate" basically affectively means, (update, insert, delete, and modify acl)

src/helpers/chains.ts
```diff
@@ -108,6 +108,15 @@ export function getBaseUrl(chainNameOrId: ChainName | number): string {
...
+/**
+ * Override the internal list of registry addresses and validator urls that will be used for Contract calls and read queries
+ * @param chainNameOrId Either the chain name or chainId.  For a list of chain names see the evm-tableland networks file
+ * @param values The values you would like to use to override the defaults.
+ *  Example: {contractAddress: "0x000deadbeef", baseUrl: "https://my.validator.mydomain.tld"}
+ * @returns void
+ */
+// TODO: It seems important to add this to the docs somewhere since it's the key
+//    to using the SDK for the non-default Validator
 export function overrideDefaults(
```
joewagner, line 120: cc: @dtbuchholz

src/helpers/config.ts
```diff
@@ -15,6 +16,17 @@ export interface AutoWaitConfig {
...
 
+export async function checkWait(
+  config: Config & Partial<AutoWaitConfig>,
+  receipt: WaitableTransactionReceipt
+): Promise<WaitableTransactionReceipt> {
+  if (config.autoWait ?? false) {
+    const waited = await receipt.wait();
+    return { ...receipt, ...waited };
+  }
+  return receipt;
```
joewagner, line 27: We had this exact block of code used in several places, so I setup a helper for it.

src/helpers/ethers.ts
```diff
@@ -39,35 +39,64 @@ export async function getOverrides({
   return opts;
 }
 
+/**
+ * RegistryReceipt is based on the TransactionReceipt type which defined by the API spec.
+ * The API v1 has a known problem where it only returns the first tableId from a transaction.
```
joewagner, line 44: see https://github.com/tablelandnetwork/go-tableland/issues/522

src/lowlevel.ts
```diff
@@ -58,6 +69,10 @@ ${padding}${carrots}`;
   },
 ];
 
+// TODO: this only works if the transaction will only be affecting a single table.
+//       I've currently got new versions of this below called execMutateMany and
+//       execCreateMany, but we might be able to combine all of these `exec` functions
+//       into one when we move to version 5.
```
joewagner, line 75: see https://github.com/tablelandnetwork/go-tableland/issues/522

src/registry/run.ts
```diff
@@ -32,21 +32,93 @@ export async function prepareRunSQL({
...
-  { statement, tableId, chainId }: RunSQLParams
+  { runnables, chainId }: MutateManyParams
 ): Promise<ContractTransaction> {
   const caller = await signer.getAddress();
   const { contract, overrides } = await getContractAndOverrides(
     signer,
     chainId
   );
-  return await contract.runSQL(caller, tableId, statement, overrides);
+  return await contract["mutate(address,(uint256,string)[])"](
```
joewagner, line 115: This function name isn't the nicest, but when we do the work to upgrade to ethers.js v6, it will go back to something like
```
await contract.mutate()
```

src/registry/utils.ts
```diff
@@ -15,8 +17,25 @@ import {
...
+ * @typedef {Object} WaitableTransactionReceipt
+ * @property {function} wait - Async function that will not return until the validator has processed tx.
+ * @property {string} name - The full table name.
+ * @property {string} prefix - The table name prefix.
+ * @property {number} chainId - The chainId of tx.
+ * @property {string} tableId - The tableId of tx.
+ * @property {string} transaction_hash - The transaction hash of tx.
+ * @property {number} block_number - The block number of tx.
+ * @property {Object} error - The first error encounntered when the Validator processed tx.
+ * @property {number} error_event_idx - The index of the event that cause the error when the Validator processed tx.
```
joewagner, line 38: Since this type is generated from the api spec, but because our api has [a bug](https://github.com/tablelandnetwork/go-tableland/issues/522), it's kind of hard to reason about what this actually looks like.
I added this here for to help with readability in the short term.  We might want to remove this comment in the future since the definition comes from a different repo and keeping this in sync with that could be prone to oversight.

src/registry/utils.ts
```diff
@@ -153,3 +180,91 @@ export async function wrapTransaction(
...
+  ): Promise<TransactionReceipt & Named & MultiEventTransaction> => {
+    const receipt = await pollTransactionReceipt(conn, params, opts);
+    if (receipt.error != null) {
+      throw new Error(receipt.error);
+    }
+    // TODO: including `name`, `prefix`, and `tableId` for back compat, will be removed next major
+    return {
+      ...receipt,
+      names,
+      name: names[0],
```
joewagner, line 250: see https://github.com/tablelandnetwork/go-tableland/issues/522

test/database.test.ts
```diff
@@ -8,8 +8,10 @@ import { Statement } from "../src/statement.js";
 
 describe("database", function () {
   this.timeout("10s");
+
+  const accounts = getAccounts();
   // Note that we're using the second account here
-  const [, wallet] = getAccounts();
+  const wallet = accounts[1];
```
joewagner, line 14: `getAccounts()` creates new instances of all of the wallets on each call.  Not a big deal, but this saves a little on memory.

test/database.test.ts
```diff
@@ -73,27 +75,107 @@ describe("database", function () {
       });
     });
 
+    test("when trying to batch create with single statement that creates 2 tables it fails", async function () {
+      const batch = db.batch([
+        db.prepare(
+          `CREATE TABLE my_table_1 (id INTEGER, name TEXT);CREATE TABLE my_table_2 (id INTEGER, name TEXT);`
```
joewagner, line 81: Not allowed to create more than one table with each statement

test/database.test.ts
```diff
@@ -73,27 +75,107 @@ describe("database", function () {
...
+      const { meta: tableMeta } = await db
+        .prepare(
+          "CREATE TABLE test_batch (id integer, name text, age integer, primary key (id));"
+        )
+        .run();
+      const tableName = tableMeta.txn?.name ?? "";
+
       const stmt = db.prepare(
-        `INSERT INTO wontwork_31337_3 (name, age) VALUES (?1, ?2)`
+        `INSERT INTO ${tableName} (name, age) VALUES (?1, ?2)`
```
joewagner, line 115: created a table and did some regular inserts to ensure things the thing under test is what we think it is.

test/database.test.ts
```diff
@@ -73,27 +75,107 @@ describe("database", function () {
...
+        .prepare(
+          "CREATE TABLE test_batch (id integer, name text, age integer, primary key (id));"
+        )
+        .run();
+      const tableName2 = tableMeta.txn?.name ?? "";
+
+      // since one sql string is touching two tables, it needs to split it into 2 runnables
+      const batch2 = db.batch([
+        db.prepare(
+          `INSERT INTO ${tableName} (name, age) VALUES ('foo', 2);INSERT INTO ${tableName2} (name, age) VALUES ('bar', 4);`
```
joewagner, line 147: The caller needs to split these into separate statements, e.g.
```
const batch2 = db.batch([
        db.prepare(`INSERT INTO ${tableName} (name, age) VALUES ('foo', 2);`),
        db.prepare(`INSERT INTO ${tableName2} (name, age) VALUES ('bar', 4);`)
]);
```

test/registry.test.ts
```diff
@@ -199,9 +221,46 @@ describe("registry", function () {
...
+        tableId: receipt.tableIds[0],
+        statement: `UPDATE test_runsql_${receipt.chainId}_${receipt.tableIds[0]} SET counter=2`,
+      });
+      const rec = await tx.wait();
+      strictEqual(typeof rec.transactionHash, "string");
+      strictEqual(rec.transactionHash.length, 66);
+    });
+  });
+
+  describe(" *deprecated* runSQL()", function () {
```
joewagner, line 261: Keeping the old tests to ensure back compat.

asutula:
Going to dig into the details tomorrow, but just making sure it is known... I believe there is code in the validator that prevents queries within a single transaction to touch multiple tables. I'm guessing that will need to change in order to support this.

joewagner:
> Thanks for the additional info @joewagner. I'm still a bit confused, but definitely am not as familiar with the situation as you are, so will defer to you in the end. I did run the equivalent of your
> 
> ```
> await db.batch([
>     db.prepare(`insert into my_table_1_11155111_123 (val) values (1);`),
>     db.prepare(`insert into my_table_2_11155111_124 (name) values ('foo');`)
> ]);
> ```
> 
> example yesterday, and it fails because the validator returns this error:
> 
> ```
> Error: BATCH_ERROR: parsing query: queries are referencing two distinct tables: users_31337_4 teams_31337_2
> ```
> 
> which I traced down to this validator code: https://github.com/tablelandnetwork/go-tableland/blob/a21d4950f50e299281380229e0eb097c91ed9cf1/pkg/parsing/impl/validator.go#L136
> 
> And of course that was using the current SDK and validator, so maybe that is different in this sdk branch.

This PR introduces the ability to do what you have listed above. However it relies heavily on the incoming changes to the Registry Contract, which are not deployed to any of the public networks yet.   I should have added a link to that PR since it's definitely key to this PR. https://github.com/tablelandnetwork/evm-tableland/pull/257
I'll add the link to the description.

If you the above command with the `latest` npm releases, it will fail because the SDK calls the Registry Contract in a way the tries to affect two tables in one event.  This PR, in combination with the Registry changes being deployed, will result in one event for each table being emitted in the transaction.  Which is something the Validator started supporting in https://github.com/tablelandnetwork/go-tableland/pull/213
Maybe we can do a voice chat?

src/database.ts
asutula: Think it's good to keep compatability.

asutula (approved):
Looks like a great update. It's a lot to take in not being more familiar with the SDK, so my approval is based on a quick review plus the conversation we had the other day where you explained all this to me.

PR #443, opened 2023-03-29
initial attempt to build out error hint scaffold

joewagner:


test/statement.test.ts
```diff
@@ -68,7 +68,7 @@ describe("statement", function () {
         (err: any) => {
           strictEqual(
             err.cause.message,
-            "error parsing statement: syntax error at position 36 near 'blurg'"
+            "error parsing statement: syntax error at position 36 near 'blurg'\n  CREATE TABLE test_run (counter blurg⚠);"
```
joewagner, line 71: @dtbuchholz I tried a few different formatting ideas, but having these errors be helpful to an application, and a terminal, and the different browser consoles is kinda restrictive.
What I landed on (so far) is inserting the unicode warning sign character.  This isn't the prettiest thing, but it's something you can find with your eyes or in code (with something like `indexOf`).
If you have any ideas or alternatives I can definitely explore those.

src/lowlevel.ts
```diff
@@ -51,6 +51,50 @@ export function errorWithCause(code: string, cause: Error): Error {
   return new Error(`${code}: ${cause.message}`, { cause });
 }
 
+const warningSymbol = String.fromCharCode(0x26a0);
+const hints = [
```
joewagner, line 55: Below is the bulk of the new feature concept.  Right now it just hints for bad syntax and missing column errors.  If this is well accepted we can continually add to hints that the need arises.

src/lowlevel.ts
```diff
@@ -106,7 +150,7 @@ export function extractColumn<T = unknown, K extends keyof T = keyof T>(
   const array = Array.isArray(values) ? values : [values];
   return array.map((row: T) => {
     if (row[colName] === undefined) {
-      throw new Error(`column not found: ${colName.toString()}`);
+      throw new Error(`no such column: ${colName.toString()}`);
```
joewagner, line 153: This kind of error can come from the Validator or the client.  This changes the client phrase to match the Validator's phrase.

src/statement.ts
```diff
@@ -80,7 +81,8 @@ export class Statement<S = unknown> {
     try {
       return bindValues(this.sql, this.parameters);
     } catch (cause: any) {
-      throw errorWithCause("BIND_ERROR", cause);
+      const hint = errorWithHint(this.sql, cause);
```
joewagner, line 84: `hint` is `cause` extended so the sql statement is in the `cause.message`.

joewagner (commented):
@dtbuchholz When you have time it would be great to get your feedback here.  This is scoped to be part of the next ship, so if this can get worked out by Friday that would be ideal.
Thanks!

test/statement.test.ts
dtbuchholz: @joewagner sorry for the delay! yes, i think this is already a great addition and would achieve what i was looking for.

the only thing i can think of that's different than ⚠ would be something like this:
```
error parsing statement: syntax error at position 36 near 'blurg'
CREATE TABLE test_run (counter blurg);
                               ^^^^^^
```
but I think that what you have is likely easier and already helpful enough—getting those `^` to line up can probs get messy.

src/lowlevel.ts
dtbuchholz: nice, this will help with error-related docs, too...more unified docs!

test/statement.test.ts
joewagner: Hmm, I like the look of the carets more than the unicode symbol....
It will probably be kinda tricky, but I'll give it a try.

joewagner: @dtbuchholz changing to use the carrots was fairly easy.  Give this another look when you get a chance.

dtbuchholz (approved):
@joewagner the carrots def look nicer and easier to identify...lgtm!

PR #448, opened 2023-04-06
chore: use github api to create signed tag

carsonfarmer:
Ensures tags are verified in the GitHub UI.


joewagner (approved):
Looks good to me!  As always with changes to workflows we will have to try it to be sure, but I don't see why this wouldn't work as expected.

PR #471, opened 2023-04-28
Fix intermittent gh action test failures

joewagner:
Recently the github actions test workflow has been failing intermittently during automated PR reviews.
This is a hard issue to track down, but the best I can tell it is a test timeout problem.   These changes introduce the ability to increase test timeouts for github actions, while leaving a normal timeout for devs running tests locally.

**NOTE:** this PR also includes a merge of main into staging

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
         run: npm run build
 
       - name: Test/Coverage 🧪
-        run: npm run coverage
+        run: TEST_TIMEOUT_FACTOR=3 npm run coverage
 
       - name: Install Playwright Browsers
         run: npx playwright install --with-deps
--- a/package.json
+++ b/package.json
@@ -107,7 +107,7 @@
     "typescript": "^4.9.4"
   },
   "dependencies": {
-    "@tableland/evm": "^4.2.0",
+    "@tableland/evm": "^4.2.1",
     "@tableland/sqlparser": "^1.0.5",
     "ethers": "^5.7.2"
   }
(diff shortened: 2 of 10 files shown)


.github/workflows/test.yml
```diff
@@ -49,7 +49,7 @@ jobs:
         run: npm run build
 
       - name: Test/Coverage 🧪
-        run: npm run coverage
+        run: TEST_TIMEOUT_FACTOR=3 npm run coverage
```
joewagner, line 52: We can triple the timeout for github actions, but leave our local tests at a reasonable time.

package.json
```diff
@@ -107,7 +107,7 @@
     "typescript": "^4.9.4"
   },
   "dependencies": {
-    "@tableland/evm": "^4.2.0",
+    "@tableland/evm": "^4.2.1",
```
joewagner, line 110: This PR includes the latest bug fix in main, which is not is staging yet, hence this diff.

test/database.test.ts
```diff
@@ -5,9 +5,10 @@ import { getAccounts } from "@tableland/local";
 import { getDefaultProvider } from "ethers";
 import { Database } from "../src/database.js";
 import { Statement } from "../src/statement.js";
+import { TEST_TIMEOUT_FACTOR } from "./setup";
 
 describe("database", function () {
-  this.timeout("10s");
+  this.timeout(TEST_TIMEOUT_FACTOR * 10000);
```
joewagner, line 11: This is the basic usage.  `TEST_TIMEOUT_FACTOR` defaults to `1` but can be set to anything greater than zero via an env var which lets us increase or reduce the timeout as needed.

joewagner (commented):
@carsonfarmer If everyone is happy with this approach, we should probably setup the same thing in the other packages that use mocha + local-tableland + github for tests.

awmuncy (approved):

PR #472, opened 2023-04-28
Ensure tableId is valid type at runtime

### LINKED ISSUE #470: [JST-23] Validator getTableById does not map chainId nor tableId
## Summary

The Validator's `getTableById` method does not work and maps `undefined` to both `chainId` and `tableId`, making the method unusable.

## Details

Set up a quick `Database` and create a `Validator`with hardcoded values:

```
const { Database, Validator, helpers } = require("@tableland/sdk");

const db = new Database({
  baseUrl: helpers.getBaseUrl(1),
});
const validator = new Validator(db.config);
const name = await validator.getTableById(1, 1);
```

When the query is made, it will error out with a message of `no chain id in path`—and you can see the `tableId` is also `undefined`(`https://tableland.network/api/v1/tables/undefined/undefined`):

```
ApiError: no chain id in path
    at fun (/Users/dtb/Desktop/hardhat-fevm/node_modules/@tableland/sdk/src/validator/client/fetcher.ts:214:15)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at getTable (/Users/dtb/Desktop/hardhat-fevm/node_modules/@tableland/sdk/src/validator/tables.ts:63:20)
    at Validator.getTableById (/Users/dtb/Desktop/hardhat-fevm/node_modules/@tableland/sdk/src/validator/index.ts:84:12)
    at main (/Users/dtb/Desktop/hardhat-fevm/scripts/read.js:14:19) {
  headers: HeadersList {
    [Symbol(headers map)]: Map(14) {
      'date' => 'Thu, 27 Apr 2023 16:26:05 GMT',
      'content-type' => 'text/plain; charset=utf-8',
      'content-length' => '34',
      'connection' => 'keep-alive',
      'access-control-allow-headers' => 'Accept, Accept-Language, Content-Type, Authorization',
      'access-control-allow-methods' => 'GET, POST, OPTIONS',
      'access-control-allow-origin' => '*',
      'trace-id' => '5020d094-5b76-46af-b426-1f169fd77766',
      'cf-cache-status' => 'DYNAMIC',
      'report-to' => '{"endpoints":[{"url":"https:\\/\\/a.nel.cloudflare.com\\/report\\/v3?s=jbrRodgQGBOwcuM9Peh%2FNXLsH%2BOD2qh8ruvBltu2%2F1ovAUQ9uGpS5MG%2BrJmVMSN6zZAVdJ883uJc2wx2nCYasWSv04TiRP1HmnSalSZpyxvPWnh12MGHj3XVXH8%2BwgWiYcOXLQ%3D%3D"}],"group":"cf-nel","max_age":604800}',
      'nel' => '{"success_fraction":0,"report_to":"cf-nel","max_age":604800}',
      'server' => 'cloudflare',
      'cf-ray' => '7be86cd6ca189e52-SJC',
      'alt-svc' => 'h3=":443"; ma=86400, h3-29=":443"; ma=86400'
    },
    [Symbol(headers map sorted)]: null
  },
  url: 'https://tableland.network/api/v1/tables/undefined/undefined',
  status: 400,
  statusText: 'Bad Request',
  data: { message: 'no chain id in path' }
}
```

## Environment

* Node 18
* @tableland/sdk ^4.2.0
* Hardhat ^2.13.0

<sub>From [SyncLinear.com](https://synclinear.com) | [JST-23](https://linear.app/tableland/issue/JST-23/validator-gettablebyid-does-not-map-chainid-nor-tableid)</sub>

joewagner:
## Overview
This makes the error message returned from an invalid call to `Validator.getTableById()` easier to understand.  

## Details
@dtbuchholz After looking into this I was reminded that the types are being generated from our API spec, so allowing tableId to be number or string or BigNumber was not a good solution.  Forcing the type to be string at runtime seems like the best path.
Interesting side note: in the process of setting this up I noticed that if you call
`validator.getTableById({chainId: 31337, tableId: 9007199254740993});`
the resulting http request is
`<api path>/tables/31337/9007199254740992`
Specifically anything bigger than Number.MAX_SAFE_INTEGER will potentially result in a request for the wrong  table.  Not a big deal because there are nowhere near that many tables.

fixes #470

### DIFF
--- a/src/validator/index.ts
+++ b/src/validator/index.ts
@@ -81,6 +81,12 @@ export class Validator {
    * @description Returns information about a single table, including schema information
    */
   async getTableById(params: TableParams, opts: Signal = {}): Promise<Table> {
+    if (
+      typeof params.chainId !== "number" ||
+      typeof params.tableId !== "string"
+    ) {
+      throw new Error("cannot get table with invalid chain or table id");
+    }
     return await getTable(this.config, params);
   }
 
--- a/test/validator.test.ts
+++ b/test/validator.test.ts
@@ -283,6 +283,36 @@ describe("validator", function () {
           return true;
         }
       );
+
+      await rejects(
+        // @ts-expect-error this is not the function signature
+        api.getTableById(31337, "1"),
+        (err: any) => {
+          strictEqual(
+            err.message,
+            "cannot get table with invalid chain or table id"
+          );
+          return true;
+        }
+      );
+
+      await rejects(
+        // params have wrong types
+        api.getTableById({
+          // @ts-expect-error we want to test that these params are being runtime type checked
+          chainId: "31337",
+          // @ts-expect-error must ensure that tableId is not a number because erc721 tokenId can be
+          // larger then JS max safe int, and string coercion will result in an incorrect value
+          tableId: 1,
+        }),
+        (err: any) => {
+          strictEqual(
+            err.message,
+            "cannot get table with invalid chain or table id"
+          );
+          return true;
+        }
+      );
     });
 
     test("when we call the tables api on a missing table", async function () {
@@ -456,7 +486,7 @@ describe("validator", function () {
   describe("rate limit", function () {
     test("when we make too many calls and get an exception", async function () {
       await rejects(
-        Promise.all(getRange(10).map(async () => await api.health())),
+        Promise.all(getRange(15).map(async () => await api.health())),
         (err: any) => {
           strictEqual(err.message, "Too Many Requests");
           return true;


dtbuchholz:
gotcha. and one day, we'll hit that max!! until then...makes sense.

joewagner:
> @joewagner i do wonder, though, if this will lead to confusion because i'd definitely assume `tableId` is a number, considering it's a numeric value to the human eye as well as its on-chain representation.
> 
> thinking about it again, maybe [`BigInt` and `.toString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt/toString) could be used to ensure spec conformity?

I can definitely understand wanting to allow bigint or number here.  The allowed type is defined in the spec https://github.com/tablelandnetwork/docs/blob/main/specs/validator/tableland-openapi-spec.yaml#L182
I can definitely find a way to work around things for typescript folks, but it feels weird to have a type defined in the spec that's different than what the SDK allows.  @carsonfarmer do you have any suggestions, or thoughts here?

awmuncy (approved):
I'd agree; the SDK should take the args in the spec. If someone has a big int, and they see it takes either a string or number, they can `toString()` it themselves, no?

joewagner:
> I'd agree; the SDK should take the args in the spec. If someone has a big int, and they see it takes either a string or number, they can `toString()` it themselves, no?

Yeah after thinking this over a bit, we should avoid deviating from the spec whenever possible.
I'll rebase this and point it at staging now so we can get it into next week's ship.

PR #483, opened 2023-05-03
upstream merge main into staging

joewagner:
We can't use the automated "update branch" github feature because of the branch protection rules.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -22,18 +22,25 @@ jobs:
             ${{ runner.os }}-build-
             ${{ runner.os }}-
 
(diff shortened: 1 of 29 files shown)


awmuncy (approved):
Hmm, annoying we can't automatically merge main into staging. Was this just a dependabot update to main? Should we point those at staging? Anyway, approved.

joewagner:
> Hmm, annoying we can't automatically merge main into staging. Was this just a dependabot update to main? Should we point those at staging? Anyway, approved.

Definitely adding some friction with the staging branch protection rules.  Seems like something could be done to make this a bit easier.

PR #491, opened 2023-05-08
Fix intermittent failing tests due to polling timeout

joewagner:
## Overview
When tests are run via github actions, the tests are failing occasionally with an error saying "the operation was aborted".  Because reproducing this failure is difficult I can't be sure if these changes really fix the problem, but it seems related to our timeout abort controller. 

## Details
In searching for the cause of the test failures I discovered 2 potential problems.  
1. using a non-default polling interval is not possible.  The default is set to 1.5 seconds, which might be too fast for some networks. This is probably not the root cause of the failing tests, but it seems good to be able to reduce the polling interval if needed.
2. The `setTimeout` that is used to enforce the maximum amount of time that should be waited, while polling, is not ever being cleared.  This means that even if the request for a transaction receipt succeeds right away, we are still eventually calling `AbortController.abort()`.  This seems potentially related to the failing tests, but again, I can't say for sure since they only happen in the github actions workflow.

### DIFF
--- a/src/helpers/await.ts
+++ b/src/helpers/await.ts
@@ -24,20 +24,23 @@ export type AsyncFunction<T> = () => Awaitable<AsyncData<T>>;
 export function getAbortSignal(
   signal?: AbortSignal,
   maxTimeout: number = 60_000
-): AbortSignal {
+): {
+  signal: AbortSignal;
+  timeoutId: ReturnType<typeof setTimeout> | undefined;
+} {
   let abortSignal: AbortSignal;
+  let timeoutId;
   if (signal == null) {
     const controller = new AbortController();
     abortSignal = controller.signal;
-    console.log(`\n\nsetting maxTimeout: ${maxTimeout}\n\n`);
-    setTimeout(function () {
-      console.log(`\n\nAborting with maxTimeout: ${maxTimeout}\n\n`);
+    // return the timeoutId so the caller can cleanup
+    timeoutId = setTimeout(function () {
       controller.abort();
     }, maxTimeout);
   } else {
     abortSignal = signal;
   }
-  return abortSignal;
+  return { signal: abortSignal, timeoutId };
 }
 
 export async function getAsyncPoller<T = unknown>(
@@ -47,7 +50,7 @@ export async function getAsyncPoller<T = unknown>(
 ): Promise<T> {
   // in order to set a timeout other than 10 seconds you need to
   // create and pass in an abort signal with a different timeout
-  const abortSignal = getAbortSignal(signal, 10_000);
+  const { signal: abortSignal, timeoutId } = getAbortSignal(signal, 10_000);
   const checkCondition = (
     resolve: (value: T) => void,
     reject: (reason?: any) => void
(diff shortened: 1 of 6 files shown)


src/statement.ts
```diff
@@ -109,6 +109,17 @@ export class Statement<S = unknown> {
...
 
+  async #execAndReturn<T = unknown>(
+    params: ExtractedStatement,
+    perfStart: number
+  ): Promise<Result<T>> {
+    const receipt = await checkWait(
+      this.config,
+      await exec(this.config, params)
+    );
+    return wrapResult<T>(receipt, performance.now() - perfStart);
```
joewagner, line 120: This isn't related to this PR, but I noticed we repeat this code more than once.

src/statement.ts
```diff
@@ -117,15 +128,15 @@ export class Statement<S = unknown> {
...
   ): Promise<Result<T>>;
   async all<T = S, K extends keyof T = keyof T>(
     colName: K,
-    opts?: Signal
+    opts?: SignalAndInterval
   ): Promise<Result<T[K]>>;
   async all<T = S, K extends keyof T = keyof T>(
     colName?: K,
-    opts: Signal = {}
+    opts: SignalAndInterval = {}
```
joewagner, line 139: I think we need to allow passing an interval to the Statement methods in order to enable setting the value to something other than the default.

src/statement.ts
```diff
@@ -198,7 +204,12 @@ export class Statement<S = unknown> {
           return extractColumn(results, colName);
         }
         default: {
-          const receipt = await exec(this.config, { type, sql, tables });
+          const receipt = await exec(this.config, {
+            ...opts,
```
joewagner, line 208: Pass the non-default interval to `exec`

src/helpers/await.ts
```diff
@@ -24,34 +24,47 @@ export type AsyncFunction<T> = () => Awaitable<AsyncData<T>>;
...
+} {
   let abortSignal: AbortSignal;
+  let timeoutId;
   if (signal == null) {
     const controller = new AbortController();
     abortSignal = controller.signal;
-    setTimeout(controller.abort.bind(controller), maxTimeout);
+    // return the timeoutId so the caller can cleanup
+    timeoutId = setTimeout(function () {
+      controller.abort();
```
joewagner, line 38: I still can't say for sure what is causing the tests to occasionally fail, but this change seems good in general.  If the request has succeeded, or the controller has already been aborted, we don't need to be waiting for this timeout and eventually calling abort.

src/helpers/await.ts
```diff
@@ -24,34 +24,47 @@ export type AsyncFunction<T> = () => Awaitable<AsyncData<T>>;
...
+  const { signal: abortSignal, timeoutId } = getAbortSignal(signal, 10_000);
   const checkCondition = (
     resolve: (value: T) => void,
     reject: (reason?: any) => void
   ): void => {
     Promise.resolve(fn())
       .then((result: AsyncData<T>) => {
         if (result.done && result.data != null) {
+          // We don't want to call `AbortController.abort()` if the call succeeded
+          clearTimeout(timeoutId);
```
joewagner, line 62: If the request succeeded, we can clear the timeout.  If timeoutId is undefined, this is a noop

awmuncy (approved):

PR #500, opened 2023-05-18
adds alter table tests

brunocalza:
## Summary

Adds some tests for ALTER TABLE statements

### DIFF
--- a/test/statement.test.ts
+++ b/test/statement.test.ts
@@ -78,6 +78,44 @@ CREATE TABLE test_run (counter blurg);
       );
     });
 
+    test("when alter table statement is valid", async function () {
+      {
+        const stmt = db.prepare(`ALTER TABLE ${tableName} ADD COLUMN a int;`);
+        const { results, meta, error } = await stmt.run();
+        deepStrictEqual(results, []);
+        strictEqual(error, undefined);
+        assert(meta.duration != null);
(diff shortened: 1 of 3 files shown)


joewagner:
@brunocalza Once I get all of the packages built and linked locally, the tests are passing for me.  Tomorrow I can do pre-releases of the needed npm packages this PR needs, then the gh test should all pass.

test/statement.test.ts
```diff
@@ -78,6 +78,44 @@ CREATE TABLE test_run (counter blurg);
...
+        const stmt = db.prepare(`ALTER TABLE ${tableName} ADD COLUMN a int;`);
+        const { results, meta, error } = await stmt.run();
+        deepStrictEqual(results, []);
+        strictEqual(error, undefined);
+        assert(meta.duration != null);
+        assert(meta.txn?.transactionHash != null);
+        strictEqual(meta.txn.name, tableName);
+
+        await meta.txn?.wait();
+      }
```
joewagner, line 92: @brunocalza This is kind of a nit, but instead of using scope blocks, we should split these into three different tests.  This takes a little longer to run, but keeps things separated in case there's a problem down the road.
If you're ok with it, since I have a working local setup, I can make that change and potentially add a test the ensures only the table owner can alter a table?

joewagner (approved):

PR #542, opened 2023-06-21
Set up SDK `aliases` feature

joewagner:
Draft PR for upcoming Aliases feature.

## Overview
The high level concept for this PR is that if a developer can provide a few config params, it will make using this SDK as an underlying driver for a SQLite ORM, or a query builder, very low friction.  For specific examples of using this feature have a look at the tests.
So far all of this has been done in a backward compatible way, but we could consider a major version bump simply because this feature enables a significantly different usage.

## Details
The basic concept is two-fold. It enables a developer to write SQL queries where the table name is specific to their configuration (does not need to be unique across all of Tableland), and the table name in the create statement is the same as in mutate statements. For example, you can do `CREATE TABLE users (first_name text);` then `INSERT INTO users (first_name) VALUES ('Bobby');`.  Any number of developers/apps/projects can have a table named "users" without conflicts, and there is no need to understand the concept of a table prefix.

The way this works is based on a new config param called `aliases`.  This param is an Object that has a `read` function and a `write` function.  `read` must return a mapping of the project scoped table name to the universally unique Tableland table name.  For example: `{ "users": "users_31337_2" }`.  `write` will be called anytime the Database instance is used to create a table.

### the read function
As mentioned above the `read` function must return a mapping of table names.  This mapping is used by our wasm sql parser via the nameMap argument.  The internals of how that works is [here](https://github.com/tablelandnetwork/wasm-sqlparser/blob/main/main.go#L157), but the TL;DR; is that the parser will replace the table names within the given statement.

### the write function
The `write` function will be called any time a change to the project might be needed.  Specifically, when a table or tables are created the function will be called with mapping of the project name (also called prefix) to the uu table name.  The developer has the option to handle this however they see fit.  
A few potential options are:
- managing a json (or yaml, or really any format) file with table names.  This seems like it would be most useful for projects were all of the tables are created during an initialization step, or the tables already exist on chain.
- managing table names via a Tableland table.  If the dev has some rules in place to manage writes to a Tableland table they could easily create a table with two columns, `project_table` and `uu_table`.  This would allow dynamically adding to, or removing from, the Project's tables.
- managing table names via a traditional database managed by the app developer, e.g. a Postgres instance.
- It's only useful for short lived sandboxes like tests, but an example in-memory mapping is setup [here.](https://github.com/tablelandnetwork/js-tableland/pull/542/files#diff-6245813f8a7143a17d54ca18a26d1d1d046aabd8512a1b70736b75994a847513R35)


src/helpers/config.ts
```diff
@@ -69,3 +78,18 @@ export async function extractChainId(conn: Config = {}): Promise<number> {
 
   return chainId;
 }
+
+export function jsonFileAliases(filepath: string): AliasesNameMap {
```
joewagner, line 82: This is really more of an example than part of the SDK, but it seems like a very common use-case and should be useful to get new devs up and running.
It's only useful in the Node.js environment.

src/helpers/parser.ts
```diff
@@ -6,18 +6,22 @@ import {
...
 } from "@tableland/sqlparser";
 import { isTestnet } from "./chains.js";
+import { type NameMapping } from "./config.js";
 
 export type { NormalizedStatement, StatementType };
 
-export async function normalize(sql: string): Promise<NormalizedStatement> {
+export async function normalize(
+  sql: string,
+  nameMap?: NameMapping
```
joewagner, line 15: The wasm parser already has `nameMap` as an optional second argument. Just need to add it here for TS to work.

src/registry/utils.ts
```diff
@@ -234,32 +234,33 @@ export async function wrapManyTransaction(
...
 
   const params = { ..._params, chainId };
+  // TODO: including `name`, `prefix`, and `tableId` for back compat, will be removed next major
+  const tableMeta = {
+    names,
+    name: names[0],
+    tableId: _params.tableIds[0],
+    prefixes,
+    prefix: prefixes[0],
+  };
```
joewagner, line 244: This is not related, just doing some cleanup/drying out.

joewagner (commented):
This is ready for a review.  I added some inline comments, and as always ping me in Discord to chat sync about any of this.
The tests are good place to get a feel for how the feature works.  In general, once you get the aliases config setup using the SDK becomes a lot more like any SQL database driver.

joewagner:
@dtbuchholz I tagged you as a reviewer, since this is a pretty dev-centric feature.  No response needed, but If you have any suggestions for improvements here feel free to chime in.

dtbuchholz:
@joewagner forsure, this is such a great feature! just to make sure i understand the current process:
1) i create a table
2) i take that table name (e.g., `mytable_31337_2_31337_3`) and manually put that in a local JSON file like `{ 'mytable': 'mytable_31337_2_31337_3' }`
3) i can make reads/writes with `mytable`, and trying to create a new `mytable` will throw

so, if i'm a dev and i want to use aliases, do i have to (1) create a JSON file mapping and then (2) handle reading that file myself? or does the SDK automatically handle reading the file...like a standard `tablelandrc.json` file that has mappings in it?

e.g., would i have to do something like the setup below where i read my JSON file and store it as `nameMap` and then instantiate?:
```ts
const db = new Database({
  signer,
  // this parameter is the core of the aliases feature
  aliases: {
    read: async function () {
      return nameMap;
    },
    write: async function (names) {
      for (const uuTableName in names) {
        nameMap[uuTableName] = names[uuTableName];
      }
    },
  },
});
```

(i can play around with it too but just some initial thoughts)

joewagner:
@dtbuchholz Definitely try it out if you have some time.  The way it's likely to work will depend a lot on the environment.  If the dev is using Tableland in node.js they can do something like:
```
const db = new Database({ signer, aliases: jsonFileAliases("my-aliases.json"), autoWait: true });
await db.prepare("create table mytbl (name text, info text);").all();
await db.prepare("insert into mytbl (name, info) values ('alias feature', 'is super cool');").all();
```
The dev would have to make sure that the `my-aliases.json` file exists, even if it's content is just `{}`.  After that there's never a need to think about the chainId or tableId.

If the dev is building something for the browser, it's a little more complicated because they have to find a way to get the name aliases.  An example might be:
```
const db = new Database({
  // Notice the browser gives us the signer
  aliases: {
    read: async function () {
      const res = doFetch("https://myapi.example.com/name-map");
      return res.json();
    }
    //  Might not need a write function if the app doesn't create tables
  }
});
```

src/statement.ts
```diff
@@ -101,12 +101,32 @@ export class Statement<S = unknown> {
...
+        : undefined;
+    const { type, statements, tables } = await normalize(
+      statementWithBindings,
+      nameMap
+    );
+
+    // TODO: this feels wrong, but we have to use the normalize function to determine the type,
+    //    then use it again if the type is create to ensure that a new table isn't created with
+    //    an existing prefix.
+    if (type === "create" && nameMap != null) {
```
asutula, line 116: `nameMap` can be `undefined`, but you are comparing to `null` here, so this will never match.

asutula (commented):
Looking good. Mostly want to discuss a potential optimization to avoid reading the aliases every time a query runs. Left a comment about it in the code.

src/database.ts
```diff
@@ -98,10 +98,17 @@ export class Database<D = unknown> {
   ): Promise<Array<Result<T>>> {
     try {
       const start = performance.now();
+      const nameMap =
+        typeof this.config.aliases?.read === "function"
+          ? await this.config.aliases?.read()
+          : undefined;
```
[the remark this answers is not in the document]
joewagner: Yeah, we've had a few issues reported where JS users get confusing run time errors.

src/database.ts
```diff
@@ -98,10 +98,17 @@ export class Database<D = unknown> {
   ): Promise<Array<Result<T>>> {
     try {
       const start = performance.now();
+      const nameMap =
+        typeof this.config.aliases?.read === "function"
+          ? await this.config.aliases?.read()
```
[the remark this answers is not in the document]
joewagner: I was doing the read each time in case there are two `Database` instances making writes. In which case you would need to have some kind of way to avoid stale caches.
The interface might be a little confusing, but if the dev wanted to optimize by putting the aliases in memory I was thinking they would do something like:
```
const nameMap = {}; // they could always do an async "fetch" here to get existing aliases
const db = new Database({
  signer,
  aliases: {
    read: async function () {
      return nameMap;
    },
    write: async function (names) {
      for (const uuTableName in names) {
        nameMap[uuTableName] = names[uuTableName];
      }
    },
  },
});
```

They would also need to build some logic to store the aliases permanently.
Just to brainstorm...
Would it be useful to enable another helper that managed an in memory Object somehow?
Maybe something like:
```
import { Database, helpers } from "@tableland/sdk":

// get the existing state however the dev wants
const aliasesObj = await doFetchToGetAliases();
const memAliases = new helpers.MemoryAliases(aliasesObj);

const db = new Database({
  signer,
  aliases: memAliases,
});

// ...later on, whenever it makes sense...
const newMap = memAliases.read();
await doPostToUpdateAliases(newMap);
```

src/helpers/config.ts
joewagner: I like it!

src/statement.ts
joewagner: When the typescript transpilation happens `!=` ends up ensuring the value is not `undefined` or `null`, which is needed for JS users I think.

asutula: Ah ok, tricky. I guess I've never thought that deeply about transpiler behaviors and JS usage when writing typescript. Would be curious to know if, in general, this type of stuff is documented somewhere?

asutula: Just to clarify, would it also be fine if you used `nameMap != undefined`? Would the transpiler still output a check for `undefined` or `null`? If so, I still think `nameMap != undefined` would be good here since it reads better in typescript and looks correct relative to our other typescript code that is in scope.

asutula (approved):

src/statement.ts
joewagner: I'm trying to rethink this with fresh eyes, and would it we better to check that `nameMap` is an Object?

RE: null vs. undefined
`== undefined` and `== null` are functionally identical.  They both check for `undefined` or `null`. I thought Typescript was transpiling `val == null` to `(val === null || val === undefined)`. But after looking in detail at the dist directory it looks like I was wrong.
Which to use feels like a style choice, and it would be nice to stay consistent.  In this package we use `== null` or `!= null` everywhere. I'm happy changing to using `undefined`, but maybe do it everywhere?  As a third option, we could also stop using [coercive comparisons](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality#comparison_with_type_conversion) all together?

asutula: Yea seems like checking that it's an object makes sense.

Oh yea, coercive types... so that does actually work as expected, ignore my original comment. In general, I've always steered away from using coercive types because they cause this type of confusion and are black magic. Plus it's nice to have `null` and `undefined` mean different things in some cases. For example, `undefined` can mean "this has never been initialized" and `null` can mean "this was set to the empty value".

PR #545, opened 2023-06-26
Allow subscribing to table changes.

joewagner:
This was a feature request that's been in limbo for about three months.  I rebased and I'll open this after some clean up.  It's a fully backward compatible change, so would be a minor bump. We can take it or leave it, but seems like we are in a good spot to decide with the start of this season.

### DIFF
--- a/src/helpers/subscribe.ts
+++ b/src/helpers/subscribe.ts
@@ -12,14 +12,6 @@ import { extractBaseUrl, type Config } from "../helpers/index.js";
 // @ts-expect-error Seems like this package isn't setup to work with modern esm + ts
 const fromEmitter = asyncGenFromEmit.default;
 
-type ContractEventTableIdMap = Record<
-  string,
-  {
-    tableIdIndex: number;
-    emit: string;
-  }
->;
-
 type ContractMap = Record<number, TablelandTables>;
 
 interface ContractEventListener {
@@ -35,6 +27,17 @@ type ListenerMap = Record<
     tableId: string;
     emitter: EventEmitter;
     contractListeners: ContractEventListener[];
+  }
+>;
+
+type ContractEventTableIdMap = Record<
+  // the key is the event name in the Solidity contract
+  string,
+  {
+    // `tableIdIndex` is the index of the event's argument that contains the tableId
+    tableIdIndex: number;
+    // `emit` is the name of the event that will be emitted by the TableEventBus instance
+    emit: string;
   }
 >;
 
@@ -139,7 +142,7 @@ export class TableEventBus {
    * Remove a listener (or iterator) based on chain and tableId
    * @param params A TableIdentifier Object. Must have `chainId` and `tableId` keys.
    */
-  removeTableListener(params: TableIdentifier): void {
+  removeListener(params: TableIdentifier): void {
     if (params == null) {
       throw new Error("must provide chainId and tableId to remove a listener");
     }
@@ -167,9 +170,9 @@ export class TableEventBus {
         const listenerObjChainId = listenerObj.chainId.toString();
 
         if (listenerObjChainId === chainId) {
-          // the chainId of the contract and the Listener Object are the same
+          // If the chainId of the contract and the Listener Object are the same
           // then we want to dig into the Listener Object and for each event that
-          // the contract is listening to we call contract.off(eventName, eventListenerFunctionInstance)
+          // the contract is listening to we remove the listener
           for (let i = 0; i < listenerObj.contractListeners.length; i++) {
             const listenerEventFunc = listenerObj.contractListeners[i];
             contract.off(
@@ -185,7 +188,7 @@ export class TableEventBus {
     // the emitter listeners and delete the entries
     for (const listener in this.listeners) {
       const l = this.listeners[listener];
-      this.removeTableListener({
+      this.removeListener({
         chainId: l.chainId,
         tableId: l.tableId,
       });
@@ -195,6 +198,7 @@ export class TableEventBus {
   async _getContract(chainId: number): Promise<TablelandTables> {
     if (this.contracts[chainId] != null) return this.contracts[chainId];
     if (this.config.signer == null) {
+      /* c8 ignore next 2 */
       throw new Error("signer information is required to get contract");
     }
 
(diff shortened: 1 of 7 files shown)


sanderpick:
nice, cool to see this landed!

sanderpick:
i'll let @carsonfarmer do the review

src/helpers/subscribe.ts
```diff
@@ -0,0 +1,208 @@
+// TODO: need to see if EventEmitter works in the browser
+import { EventEmitter } from "node:events";
```
carsonfarmer, line 2: There are browser based alternatives if not. There might even be some that work in both.

src/helpers/subscribe.ts
```diff
@@ -0,0 +1,208 @@
...
+    chainId: number;
+    tableId: string;
+    emitter: EventEmitter;
+  }
+>;
+
+/**
+ * List of the Registry Contract events that will be emitted from the TableEventBus.
+ */
+// TODO: this is an awkward way to keep track of the index of the Events' tableId index
```
carsonfarmer, line 34: Is it? Seems ok to me?

src/helpers/subscribe.ts
```diff
@@ -0,0 +1,208 @@
...
+  /**
+   * Start listening to the Registry Contract for events that are associated
+   * with a given table.
+   * There's only ever one "listener" for a table, but the emitter that
+   * Contract listener has can have as many event listeners as the environment
+   * supports.
+   * @param tableName The full name of table that you want to listen for
+   * changes to.
+   */
+  async addTableListener(tableName: string): Promise<EventEmitter> {
```
carsonfarmer, line 85: Why `addTableListener` versus just `addListener`? Just because name clash with classic listener adding api?

src/helpers/subscribe.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  _attachEmitter(
+    contract: TablelandTables,
+    emitter: EventEmitter,
+    tableIdentifier: TableIdentifier
+  ): void {
+    const { tableId, chainId } = tableIdentifier;
+    for (const key in contractEvents) {
+      const eve = contractEvents[key];
+      // TODO: might have to put the listener function in memory so we can remove it.
```
carsonfarmer, line 178: Should we handle these TODOs?

test/subscribe.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    deepStrictEqual(results, []);
+    strictEqual(error, undefined);
+    match(tableName, /^test_table_subscribe_31337_\d+$/);
+
+    await meta.txn?.wait();
+  });
+
+  describe("addTableListener()", function () {
+    const eventBus = new TableEventBus(db.config);
```
carsonfarmer, line 33: This is pretty cool, I like this API design @joewagner!

joewagner:
> i'll let @carsonfarmer do the review

Sweet, sounds good!
I have another PR incoming that has more tests, some clean up, and the addition of an alternative async/await api

carsonfarmer (commented):
I think it is worth finishing this up and taking it out of draft mode!

src/helpers/subscribe.ts
joewagner: Definitely!  I've got a big set of changes that should handle these.  Should have it pushed in a few hours.

src/helpers/subscribe.ts
joewagner: I like it.  Not sure what my original thinking was, but I'll update it.

src/registry/contract.ts
```diff
@@ -23,7 +23,7 @@ export interface TableIdentifier {
   tableId: string;
 }
 
-async function getTableIdentifier(
+export async function getTableIdentifier(
```
joewagner, line 26: Need to export so this can be re-used in the subscribe helper.

test/validator.test.ts
```diff
@@ -114,45 +114,44 @@ describe("validator", function () {
...
           transactionHash,
         });
         match(receipt.error!, /.*msg: NOT NULL constraint failed:.*/);
       });
-      this.afterEach(() => {
-        db.config.autoWait = true;
-      });
-    });
 
-    describe("without autoWait", function () {
```
joewagner, line 143: This is not related to the PR, but I noticed that there seems to have been a copy-pasta gone wrong here.  There was an identical but separate `describe` for each of these tests, and two of the tests looked like they were testing the exact same thing.

joewagner (commented):
@carsonfarmer This is ready for a review.  Happy to jump in a chat if you want to run through it sync.

src/helpers/subscribe.ts
```diff
@@ -0,0 +1,281 @@
...
+import { type TablelandTables } from "@tableland/evm";
+import { pollTransactionReceipt } from "../validator/receipt.js";
+import {
+  getTableIdentifier,
+  getContractAndOverrides,
+  type TableIdentifier,
+} from "../registry/contract.js";
+import { extractBaseUrl, type Config } from "../helpers/index.js";
+
+// @ts-expect-error Seems like this package isn't setup to work with modern esm + ts
```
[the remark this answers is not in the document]
carsonfarmer: What about this one: https://github.com/rolftimmermans/event-iterator
Otherwise, I'm sure its fine?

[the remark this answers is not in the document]
carsonfarmer: By the way, this is a pretty useful resource: https://github.com/alanshaw/it-awesome

test/validator.test.ts
carsonfarmer: Huh, maybe a leftover merge conflict?

carsonfarmer (approved):
This looks really cool @joewagner! I love how it handles everything client-side. This is quite a clever solution to something that is actually quite tricky to deal with!

src/helpers/subscribe.ts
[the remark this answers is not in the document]
joewagner: I explored using the `event-iterator` package but it seems like it will be a little complicated to wrap an existing node.js EventEmitter instance.  Which feels odd since the description makes it sound like it would be the primary use-case, but all the examples are for streams, or browser events.
The async-generators package works as a one-liner, maybe we stick with that unless an issue arrises?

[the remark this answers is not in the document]
carsonfarmer: Yeh sure!

PR #562, opened 2023-07-18
fix: allow insert with subselect

dtbuchholz:
## Summary

Adds logic to parse a statement and validate if the query is mutating a single table while including a subselect statement.

## Details

The core logic change uses a regex `/select\s+.*?\s+from\s+(\w+)/gi` to check each individual statement within a `batch` call. It looks for a SELECT FROM and grabs the ensuing table name. Then, it checks if this table name exists in the `tables` names array returned from the SQL parser's `normalize` method. If so, "ignore" this table by filtering it out of the array.

If the `tables` array's names only affect a single table, the statement can be executed; else, more than one table is being mutated, and it'll throw.

## How it was tested

Added two tests:
- `when batching single statement, including subselect, affecting two tables throws an error`
- `when batching mutations works with a subselect`

## Checklist:

### DIFF
--- a/src/database.ts
+++ b/src/database.ts
@@ -235,9 +235,31 @@ async function normalizedToRunnables(
     );
   }
   if (normalized.tables.length > 1) {
-    throw new Error(
-      "each statement can only touch one table. try batching statements based on the table they mutate."
-    );
+    let isMutatingMultiple = true;
+    for (const stmt of normalized.statements) {
+      // re-normalize so we can be sure we've isolated each statement and its tableId
+      const norm = await normalize(stmt);
+
+      if (norm.tables.length > 1) {
+        // check if one of the tables is *not* mutating but from a subselect
+        // look for "select", find "from", and get the table name
+        const regex = /select\s+.*?\s+from\s+(\w+)/gi;
+        const tableNames = [...stmt.matchAll(regex)].map((match) => match[1]);
+        // check if these tables are in the normalized table names
+        // if so, filter them out (i.e., they are not being mutated)
+        const filteredTables = norm.tables.filter(
+          (tableName) => !tableNames.includes(tableName)
+        );
+        // if the filtered tables are greater than 1, then there are two
+        // tables being mutated in a single statement, which is not allowed
+        isMutatingMultiple = filteredTables.length > 1;
+      }
+    }
+    if (isMutatingMultiple) {
+      throw new Error(
+        "each statement can only touch one table. try batching statements based on the table they mutate."
+      );
+    }
   }
 
   const { tableId } = await validateTableName(normalized.tables[0]);
--- a/test/database.test.ts
+++ b/test/database.test.ts
@@ -142,7 +142,7 @@ describe("database", function () {
       });
     });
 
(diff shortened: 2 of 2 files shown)


src/database.ts
```diff
@@ -235,9 +235,31 @@ async function normalizedToRunnables(
     );
   }
   if (normalized.tables.length > 1) {
```
dtbuchholz, line 237: here, we check if there are > 1 tables in the normalized statement. if so, we will:
- renormlize each statement
- check for a SELECT FROM statement and get the table name
- filter the `norm.tables` array and remove this table from the array, stored as `filteredTables`
- if the length of `filteredTables` is > 1, that means there's more than one table trying to be mutated, which throws

the one callout: as noted [here](https://linear.app/tableland/issue/JSTC-34#comment-b6c4917e), we **don't** want to allow a query like this, shown using pg:
```
with ins as (
    insert into table1 (userid, email)
    values(12,'example@gmail.com') 
)
insert into table2 (userid, name, address)
values(12, 'Joe', '123 test Dr')
```
i could be wrong, but i don't think there's a way to do this in SQLite without passing multiple queries in a single runnable—i.e., `INSERT INTO table1 ...; INSERT INTO table2 ...;` will throw. so, i _think_ the `filteredTables` logic to remove a table should work. in other words, if the pg query was valid in SQLite, this logic would be flawed.

--

also, there could be more efficient ways to do this—open to suggestions.

test/database.test.ts
```diff
@@ -175,6 +175,42 @@ describe("database", function () {
...
+      const stmt = db.prepare(
+        `insert into ${tableToSubselect} (address) values (?1);insert into ${tableToMutate} (data) select ?2 from ${tableToSubselect} where address=?1;`
+      );
+      const batch = db.batch([stmt.bind(signer.address, "test")]);
+
+      // expect an error due to the individual statements touching two tables
+      await rejects(batch, (err: any) => {
+        strictEqual(
+          err.cause.message,
+          `parsing query: queries are referencing two distinct tables: ${tableToSubselect} ${tableToMutate}`
```
dtbuchholz, line 202: noting that this error is thrown instead of `each statement can only touch one table. try batching statements based on the table they mutate.`

joewagner:
@dtbuchholz In general this looks good. Your branch is not starting off the tip of main, can you either rebase it or merge main into it?

awmuncy (approved):
Reviewed this one in the course of reviewing the CLI. Looks good to me!

dtbuchholz:
cool, thanks! as far as merging, since this is a bug fix, i can merge into main?

src/database.ts
joewagner: I spent some time trying to figure out how to do something like the above pg query with sqlite and I can't come up with anything.

joewagner:
> cool, thanks! as far as merging, since this is a bug fix, i can merge into main?

Yeah, merge away.  After the merge you can also publish via the Publish action.  Looks like this will be 4.4.1

joewagner:
@dtbuchholz Hope it's ok, I'm gonna merge and publish so I can get this into the local tableland release today.

Issue #568, opened 2023-07-20
[JST-32] aliases helper needs the "node:fs" package, but this breaks browser builds


joewagner:
Consider removing the helper?

<sub>From [SyncLinear.com](https://synclinear.com) | [JST-32](https://linear.app/tableland/issue/JST-32/aliases-helper-needs-the-nodefs-package-but-this-breaks-browser-builds)</sub>

PR #569, opened 2023-07-20
Use dynamic import of "fs" module to enable webpack builds

joewagner:
@asutula or @awmuncy This fixes the setup I used to reproduce #568, but can one of you take a look at this and double check that it fixes your issue using the SDK in the Studio?

### DIFF
--- a/src/helpers/config.ts
+++ b/src/helpers/config.ts
@@ -1,4 +1,3 @@
-import fs from "node:fs";
 import { type WaitableTransactionReceipt } from "../registry/utils.js";
 import { type ChainName, getBaseUrl } from "./chains.js";
 import { type Signer, type ExternalProvider, getSigner } from "./ethers.js";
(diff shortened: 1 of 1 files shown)


src/helpers/config.ts
```diff
@@ -79,21 +78,37 @@ export async function extractChainId(conn: Config = {}): Promise<number> {
...
 
+// TODO: next major we should remove the jsonFileAliases helper and expose it
+//    in a different package since it doesn't work in the browser.
+const getFsModule = (function () {
+  let fs: any;
+  return async function () {
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    if (fs) return fs;
+
+    fs = await import(/* webpackIgnore: true */ "fs");
```
joewagner, line 99: In addition to not importing "fs" unless the dev is using the helper, I also have to tell webpack to ignore this to get the build to work.

src/helpers/config.ts
```diff
@@ -1,4 +1,3 @@
-import fs from "node:fs";
```
joewagner, line 1: Details on the difference between "node:fs" and "fs" are here: https://nodejs.org/api/modules.html#core-modules
TL;DR; the `node:` prefix ignore the require cache.  In our case we don't want to do that because the SDK consumer may want to be able to leverage the require cache.

awmuncy (approved):
Pulled this in using NPM link, and it is all working great. LGTM!

joewagner:
> Pulled this in using NPM link, and it is all working great. LGTM!

Excellent!  This fix is released as 4.4.2

asutula:
Nice, good work.

PR #575, opened 2023-07-28
Setup org wide lint, prettier, test standards

joewagner:
As part of the move to a monorepo we have to make a coordinated change from this repo to https://github.com/tablelandnetwork/tableland-js.
The changes in this PR enable the `@tableland/sdk` package to be linted, prettiered, and tested in the same way as `@tableland/cli` and `@tableland/local`.


joewagner:
related to:
https://github.com/tablelandnetwork/js-tableland-cli/pull/406
https://github.com/tablelandnetwork/local-tableland/pull/440

lint.tsconfig.json
```diff
@@ -1,4 +1,16 @@
 {
-  "extends": "./tsconfig.json",
```
awmuncy, line 2: Might it cause problems down the road that linting and the project's default tsconfig aren't tied together at all? I.e. tests are succeeding because of one tsconfig configuration, whereas the actual library fails because it doesn't have the same constraints?

awmuncy (approved):
LGTM! Left one question inline, but it isn't a blocker.

lint.tsconfig.json
joewagner: Great point, pretty much all of this is done in a way that's trying to get all the repo's linting in the same place.  I tried to leave the build steps mostly unchanged.  I removed this here to accomplish that, but when we move to the monorepo we could try to setup a shared tsconfig

joewagner:
@awmuncy I had to make one small change, can you give this another 👍 ?

awmuncy (approved):

PR #576, opened 2023-07-29
fix abort controller handling

sanderpick:
## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules


src/database.ts
```diff
@@ -92,8 +92,8 @@ export class Database<D = unknown> {
   //    D1-ORM is a good example: https://github.com/Interactions-as-a-Service/d1-orm/
   async batch<T = D>(
     statements: Statement[],
-    opts: Signal = {}
-    // reads returns an Array with legnth equal to the number of batched statements,
+    controller?: PollingController
```
sanderpick, line 95: Couple things here:
1. This wasn't really an "options" object, it was a specific type of object. I usually think of "options" as an object that mixes different type of options, e.g., color, maxHeight, etc. I changed the param name to reflect.
2. The value of `opts` was `Signal`, but this method may call mutable methods that take polling options.

src/database.ts
```diff
@@ -124,7 +124,9 @@ export class Database<D = unknown> {
       // and return an Array of the query results.
       if (type === "read") {
         return await Promise.all(
-          statements.map(async (stmt) => await stmt.all<T>(undefined, opts))
+          statements.map(
+            async (stmt) => await stmt.all<T>(undefined, controller)
```
sanderpick, line 128: I'm guessing this won't work... the timeout would be cancelled after the first result is returned. Sound right? This might need some special handling... ie, create a new controller for each statement (clone of `controller`) and control each of those with the single user-provider one. I can take a stab if sounds right.

src/helpers/await.ts
```diff
@@ -1,17 +1,19 @@
...
 }
 
 export interface Interval {
-  interval?: number;
+  interval: number;
+  cancel: () => void;
 }
 
-export type SignalAndInterval = Signal & Interval;
+export type PollingController = Signal & Interval;
```
sanderpick, line 13: Changing this to a name that more closely describes what it does. I had to dig into the source code to figure this out when I first saw this.

src/helpers/await.ts
```diff
@@ -21,36 +23,41 @@ export interface AsyncData<T> {
...
-  timeoutId: ReturnType<typeof setTimeout> | undefined;
-} {
-  let abortSignal: AbortSignal;
-  let timeoutId;
-  if (signal == null) {
-    const controller = new AbortController();
-    abortSignal = controller.signal;
-    // return the timeoutId so the caller can cleanup
-    timeoutId = setTimeout(function () {
+export function createSignal(): Signal {
```
sanderpick, line 26: Helper method for use with read-only APIs

src/helpers/await.ts
```diff
@@ -21,36 +23,41 @@ export interface AsyncData<T> {
...
-    }, maxTimeout);
-  } else {
-    abortSignal = signal;
-  }
-  return { signal: abortSignal, timeoutId };
+    },
+  };
+}
+
+export function createPollingController(
```
sanderpick, line 36: Helper method for mutating APIs.

src/helpers/await.ts
```diff
@@ -21,36 +23,41 @@ export interface AsyncData<T> {
...
+  timeout: number = 60_000,
+  pollingInterval: number = 1500
+): PollingController {
+  const controller = new AbortController();
+  const timeoutId = setTimeout(function () {
+    controller.abort();
+  }, timeout);
+  return {
```
sanderpick, line 44: This object doubles as an options object for fetch and as a controller for the user to call `abort` on.

src/helpers/config.ts
```diff
@@ -26,10 +27,11 @@ export interface AliasesNameMap {
 
 export async function checkWait(
   config: Config & Partial<AutoWaitConfig>,
-  receipt: WaitableTransactionReceipt
+  receipt: WaitableTransactionReceipt,
+  controller?: PollingController
 ): Promise<WaitableTransactionReceipt> {
   if (config.autoWait ?? false) {
-    const waited = await receipt.wait();
+    const waited = await receipt.wait(controller);
```
sanderpick, line 34: This is the actual fix for the polling handling. `wait` was being called without any options / controller.

src/lowlevel.ts
```diff
@@ -252,10 +252,10 @@ function catchNotFound(err: unknown): [] {
 export async function queryRaw<T = unknown>(
   config: ReadConfig,
   statement: string,
-  opts: Signal = {}
+  signal?: Signal
```
sanderpick, line 255: Renamed. As mentioned above for controller, this isn't really an "options" object.

src/statement.ts
```diff
@@ -191,22 +196,22 @@ export class Statement<S = unknown> {
...
   ): Promise<T>;
   async first<T = S, K extends keyof T = keyof T>(
     colName: K,
-    opts?: SignalAndInterval
+    controller?: PollingController
   ): Promise<T[K] | null>;
   async first<T = S, K extends keyof T = keyof T>(
     colName?: K,
-    opts: SignalAndInterval = {}
+    controller?: PollingController
```
sanderpick, line 214: Is there any reason this was not optional and was instead an empty object by default. I made this change in a bunch of places.

src/statement.ts
```diff
@@ -216,19 +221,21 @@ export class Statement<S = unknown> {
...
-            tables,
-          });
+          await this.#waitExec(
+            {
+              type,
+              sql,
+              tables,
+            },
+            controller
+          );
```
sanderpick, line 238: The old `opts` var was being injected into the statement object, never to be seen again. Pulling the controller out into its own var makes it easier to see what's going on.

sanderpick (commented):
@joewagner @carsonfarmer here's a draft fix for the polling behavior. Let me know what you think.

sanderpick:
Related, after poking around in here, wouldn't it be cleaner to use a traditional options object on the statement api that included the `colName` for filtering and the "polling controller", ie, something like https://github.com/tablelandnetwork/js-tableland/commit/ef8f74f7acb51709d02f1c3dada2f69a21b55dd1 ? 

Although, maybe this is part of the codebase that has to conform to D1?

src/statement.ts
carsonfarmer: Providing a default value _makes_ it effectively optional. If you look at the typescript signature, it includes the `?`. That way, you don't have to double check that a) opts is provided, and then b) if a given property on opts is provided. With chained ? this isn't as much of an issue, but it seemed to lead to cleaner code.

src/lowlevel.ts
carsonfarmer: We left it as options in case we wanted to add additional opts to it at a later date.

sanderpick:
> Yeh those are D1 APIs, please don't change without ensuring they are backward compatible with 3rd party libs.

K, do you have a ref to the API that you're conforming to here? I see a mention of D1 on the `Database` object, but nothing here.

src/lowlevel.ts
sanderpick: I see. So you would have unioned `Signal` with some other type in the future? Are you suggesting I change it back?

src/statement.ts
carsonfarmer: Yeh that's true, with those chained `?` it is pretty clean. Ok!

carsonfarmer:
Yeh, we should actually have a reference to its types. IIRC there was an issue a while back where we couldn't just do "implements" or "extends" because the D1 types didn't actually match the response you would get in practice. But they come from here: https://github.com/cloudflare/workerd/blob/main/types/defines/d1.d.ts and we can probably use the types from `@cloudflare/worker-types` now? Note that _differences_ to these types are fine as long as they are backward compatible.

src/lowlevel.ts
carsonfarmer: Yes we could have, but naw it is probably fine. Worst case is that we have to change the parameter name in the future. But since it is already an _object_, it'll be a pretty safe change.

sanderpick:
Cool, thanks! I see `colName` here: https://github.com/cloudflare/workerd/blob/main/types/defines/d1.d.ts#L26. Is that what you're looking at? Maybe it's just on `first` now in their API? btw, I don't have a strong pref here, it just felt weird to type `undefined` in the `all` method to be able to use the custom timeout.

joewagner:
@sanderpick This looks like great improvements in general.  I agree the current API is very confusing, and turns out broken.
I notice you're pointing this PR at `main` and I was hoping to stop using this repo this week. However, in order to do that we need to get all the repos aligned on linting, prettier, and tests.  I don't think there will be much issue here since the monorepo standards are mostly all being taken from this repo, but can you point this PR at the `joe/lint` branch?

sanderpick:
@joewagner Absolutely. Just wanted to get a first take on this. I'll try to get that done today.

